PR #8, opened 2020-07-23
Add linting for Javascript

zstewar1:


zstewar1:
This should help catch invalid javascript in pull requests.

PR #13, opened 2020-07-31
Load interests

Leo2452:
Save interests for a user and retrieve in another page, which loads recommendations on a map based on those interests.


byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,33 @@ private String formatLocation(PlaceType location) {
...
   }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query allUsers = new Query("sample");
+    PreparedQuery preparedUsers = datastore.prepare(allUsers);
+    for(Entity enteredUser: preparedUsers.asIterable()) {
+      long id = (long) enteredUser.getProperty("id");
+      if(id == 2452) {
```
landondare, line 71: Same comment as above. You can limit your query to just the object with this specific id

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,33 @@ private String formatLocation(PlaceType location) {
...
+        String interestsAsString = request.getParameter("data");
+        addInterestsForUser(interestsAsString, enteredUser, datastore);
+      }
+    }
+  }
+
+  /**
+   * Creates a list of interests from the JSON string passed in and adds it
```
landondare, line 79: Good documentation!

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+if(document.readyState === 'loading') {
+  document.addEventListener('DOMContentLoaded', initialize);
+} else {
+  initialize();
+}
+
+let map;
+let service;
+let randomLocation;
+let RADIUS = 1000;
```
landondare, line 26: `const RADIUS = 1000;`

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+/* global google, marker */
+
+if(document.readyState === 'loading') {
+  document.addEventListener('DOMContentLoaded', initialize);
+} else {
+  initialize();
+}
+
+let map;
+let service;
```
landondare, line 24: A more specific variable name would be good here especially since there are multiple services used by the application. Something like `placesService` would be better

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Checks the response from PlacesService and creates markers
+ * on the locations found from the request.
+ * @param {Array results} contains the results of the request.
+ * @param {PlacesServiceStatus status} contains the service status
+ * of the request.
+ */
+function callback(results, status) {
```
landondare, line 70: Please use a more descriptive name related to what this function does. E.g. something like `addMarkers()`.

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Servlet that loads places of interest from
+ * a specified user, previously configured.
+ */
+@WebServlet("/api/generator")
+public class GeneratorServlet extends HttpServlet {
+
+  private final Gson gson = new Gson();
+  private ArrayList<String> interestsSelected = new ArrayList<>();
```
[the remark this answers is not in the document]
Leo2452: I assign this variable to the list of interests for a user's Entity within the datastore. It looks for the user and then stores the information in this list, which is then converted to a json and sent to front end

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query person = new Query("sample");
```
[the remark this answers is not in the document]
Leo2452: This sample entity type is a user for now. It currently finds the user, checks the id and if a specific one is found, then it gets their interests.

I'll rename it to user for clarification. I'll also try to add some fake objects and retrieve those instead of entities for the time being

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+    PreparedQuery pq = datastore.prepare(person);
+    for(Entity enteredUser: pq.asIterable()) {
```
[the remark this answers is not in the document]
Leo2452: There should only be one user per id. This implementation currently cycles through all users and their entries, checking their id with a temp hard-coded value `2452`. I'm working on configuring the ID stuff and properly-saving/retrieving in a separate branch - I'll add a comment to the function to clarify.

byway/src/main/webapp/interests.html
```diff
@@ -10,5 +10,9 @@
     <div id="header">Some content</div>
     <p> Scroll through and select some interests</p>
     <div id="interests"></div>
+    <form id="interests-form" action="/generator.html">
```
j-vent, line 13: I think it's standard to specify the method (get or post) when using `form` tags: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+    for (let i = 0; i < results.length; i++) {
+      if(fitsInRadius(results[i], RADIUS)) {
+        createMarker(results[i]);
+        numRecommendations++;
+      }
+      if(numRecommendations == maxRecommendations) {
+        break;
+      }
+    }
+  } else {
```
j-vent, line 83: Might be good to display the status code as well in your error message.

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+public class GeneratorServlet extends HttpServlet {
+
+  private final Gson gson = new Gson();
+  private ArrayList<String> interestsSelected = new ArrayList<>();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query person = new Query("sample");
+    PreparedQuery pq = datastore.prepare(person);
```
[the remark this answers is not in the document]
Leo2452: Got it, I'll add it - thanks

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+ * on the locations found from the request.
+ * @param {Array results} contains the results of the request.
+ * @param {PlacesServiceStatus status} contains the service status
+ * of the request.
+ */
+function addRecommendations(results, status) {
+  if (status == google.maps.places.PlacesServiceStatus.OK) {
+    let maxRecommendations = 3;
+    let numRecommendations = 0;
+    for (let i = 0; i < results.length; i++) {
```
renau06, line 74: could you loop through results using foreach, so you dont have to use i to access the result?

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,112 @@
...
+if(document.readyState === 'loading') {
+  document.addEventListener('DOMContentLoaded', initialize);
+} else {
+  initialize();
+}
+
+let map;
+let placesService;
+let randomLocation;
+const RADIUS = 1000;
```
renau06, line 26: maybe document what this is? like if it is 1000 feet, meters, etc

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.google.maps;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.maps.model.PlaceType;
+import com.google.gson.Gson;
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/pom.xml
```diff
@@ -34,6 +34,11 @@
       <artifactId>google-maps-services</artifactId>
       <version>0.14.0</version>
     </dependency>
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/GeneratorServlet.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query person = new Query("sample");
+    PreparedQuery pq = datastore.prepare(person);
+    for(Entity enteredUser: pq.asIterable()) {
+      long id = (long) enteredUser.getProperty("id");
+      if(id == 2452) {
+        Entity currentUser = enteredUser;
+        interestsSelected = (ArrayList) currentUser.getProperty("interests");
```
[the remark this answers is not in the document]
Leo2452: Yeah I could have, I forgot to switch it on this branch when I fixed on a different one. Also, I modified this part to use filters instead, which avoids unnecessarily making a new variable like here - fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -14,6 +14,11 @@
...
 package com.google.maps;
 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
 import com.google.maps.model.LatLng;
 import com.google.maps.model.PlaceType;
 import com.google.gson.Gson;
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,33 @@ private String formatLocation(PlaceType location) {
     place = place.replace('_', ' ');
```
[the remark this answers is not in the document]
Leo2452: This servlet focuses on getting the placeTypes that exist within the API and formatting them, such as `Convenience store` or `Park`. It was just a small detail since it loaded them as `convenience-store` and `park`. This function isn't used for loading the recommendations list, only the places to make them as interests for the buttons. There was one exception that I found, `Rv park` which should need the v capitalized, but I was going to filter through the interests at a later date to remove other placeTypes so I'll focus on that in a future PR

byway/src/main/webapp/interests.html
Leo2452: I specify its method and other params in the `track-interests.js` file (see `loadForm`). I did this since I had to fetch per submit, so I added an event listener and handled it in that function

byway/src/main/webapp/recommendations.js
Leo2452: Fixed

byway/src/main/webapp/recommendations.js
Leo2452: I initially did use `.forEach`, but i also had to use the `i` variable to keep track of the current recommendations loaded to limit the results it showed with `maxRecommendations`, or else it could flood the map with a lot of markers. I had to keep track of the `i` counter, so I just used a standard for loop instead

Leo2452: Also, this is a temporary solution, since the user might want to load more recommendations than what we show at first. I'll add that into function's comments and set it aside as a TODO

byway/src/main/webapp/recommendations.js
Leo2452: Good point, will fix

byway/src/main/webapp/recommendations.js

byway/src/main/webapp/recommendations.js

byway/src/main/webapp/recommendations.js

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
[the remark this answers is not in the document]
j-vent: Oh I see, my bad.

PR #14, opened 2020-07-31
RouteSelection

j-vent:
Selection functionality for the third page. I am currently using hardcoded values for the recommendations, which will be replaced with values from a list in a Trip entity. As such, refreshing the page will load the same hardcoded places onto the list. However, the list of selected places is routinely updated in both the local copy in the js and the datastore. 

This is a rough implementation, so any suggestions to refactor the code are appreciated.


.gitignore
```diff
@@ -1,116 +1,2 @@
...
-# DynamoDB Local files
-.dynamodb/
-
-# TernJS port file
-.tern-port
-
-# Stores VSCode versions used for testing VSCode extensions
-.vscode-test
-
-# End of https://www.toptal.com/developers/gitignore/api/node
```
Leo2452, line 116: Many files seem to be deleted - maybe rebase?

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
+// Copyright 2019 Google LLC
+//
```
Leo2452, line 2: Consider moving these functions into a separate js file, or renaming this file to be more specific to the page you're using them on.

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Adds a random greeting to the page.
+ */
+function addRandomGreeting() {
```
Leo2452, line 18: Remove functions that are unused, especially ones that were left from the default `script.js`

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+  var directionsRenderer = new google.maps.DirectionsRenderer();
+  var start = new google.maps.LatLng(37.7699298, -122.4469157);
+  var end = new google.maps.LatLng(37.7683909618184, -122.51089453697205);
+  var mapOptions = {
+    zoom: 14,
+    center: start
+  }
+  var map = new google.maps.Map(document.getElementById('map'), mapOptions);
+  directionsRenderer.setMap(map);
+     document.getElementById("route").addEventListener("click", function() {
```
Leo2452, line 41: Align this LOC with the rest of the function's body

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+  var request = {
+      origin:  start,
+      destination: end,
+      travelMode: 'DRIVING'
+  };
+  directionsService.route(request, function(response, status) {
+    if (status == 'OK') {
+      directionsRenderer.setDirections(response);
+    }
+  });
```
Leo2452, line 57: Add an `else` condition in case the service doesn't work. Consider displaying the status as well

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function getStops(){
+  console.log("get stops!");
+  const stopList = document.getElementById('stop-list');
+  
+  fetch('/stop')
+  .then(response => response.json())
+  .then((stops) => {
+    stopList.innerHTML = ""; // clear list
```
Leo2452, line 67: Prefer using `innerText` over `innerHTML`

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+    for(let i = 0; i < stops.length; i++) {
+        const btn = document.createElement('button');
+        btn.id = `stopList${i}`;
+        // btn.className = 'list-group-item list-group-item-action';
+        btn.innerHTML = stops[i];
+        btn.addEventListener("click", function() {
+          console.log("delete me!");
+          console.log(stops[i]);
```
Leo2452, line 75: Consider adding more functionality to this event listener. Is this the part that moves the button/its content into a separate div, ie when displaying the route as user makes it?

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function deleteFromRecs(stop){
+    console.log("delete from recs");
+    const params = new URLSearchParams();
+    params.append("toDelete", stop);
+    fetch('/data', {method: 'POST', body: params})
+    .then(() => getRecs());
+     const recList = document.getElementById('rec-list');
+    recList.innerHTML = ""; // clear list
```
Leo2452, line 98: Spacing on line 97. Also, is your intent to clear `recList` after `.then(() => getRecs())`? Consider clearing the list and then loading with fetch. Or, consider solely removing one rec rather than clearing them all, then loading the remaining `n-1` elements from the server.

Leo2452: It seems you clear it inside `getRecs()`, so remove this line to prevent the same action being made.

byway/src/main/webapp/script.js
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function getRecs() {
+ console.log("get recs!")
+ const recList = document.getElementById('rec-list');
+  fetch('/data')
```
Leo2452, line 104: Try to rename your servlet route from `/data` to something else like `/recs` to be more specific. Also, try to switch all `innerHTML` into `innerText`.

byway/src/main/webapp/routepage.js
```diff
@@ -12,12 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// copies of recommendations and selected stops, stored as arrays
+// for synchronous updating
+let recs = [];
+let stops = [];
+
+let l = load;
```
Leo2452, line 20: Rename to make more specific. Not sure what this is used for.

byway/src/main/webapp/routepage.js
```diff
@@ -12,12 +12,26 @@
...
+
 if (document.readyState === 'loading') {  // Loading hasn't finished yet
-  document.addEventListener('DOMContentLoaded', getRecs);
+  document.addEventListener('DOMContentLoaded', load);
 } else {  // `DOMContentLoaded` has already fired
+  l();
```
Leo2452, line 25: I'm not too sure how callback functions work, but i think you could call `load()` here instead, if that was your reason for using the `l` variable above.

byway/src/main/webapp/routepage.js
```diff
@@ -29,11 +43,12 @@ function initMap() {
   }
   let map = new google.maps.Map(document.getElementById('map'), mapOptions);
   directionsRenderer.setMap(map);
-    document.getElementById("route").addEventListener("click", function() {
+document.getElementById("route").addEventListener("click", function() {
```
Leo2452, line 46: Spacing still seems to be off

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +64,135 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
   });
 }
 
+/** Get the new list of stops from datastore onload */
+function getStops(){
+  const stopList = document.getElementById('stop-list');
+  if(stopList != null){
+        stopList.innerText = ""; // clear list
+    }
```
Leo2452, line 72: would the stopList ever be null in any situation?

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +64,135 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  fetch('/api/stop')
+  .then(response => response.json())
+  .then((stops) => {
+    stops.forEach((stop)=>{
+      var btn = document.createElement('button');
+      btn.id = stop.id;
+      btn.innerText = stop.placename;
+      btn.setAttribute("class", "btn rec-btn");
```
Leo2452, line 80: Consider using `btn.className` instead of setting the attribute. I was told by Zach we should prefer accessing attributes this way instead. Also, you might want to use `btn.classList.add("btn rec-btn")` instead to append classes to the button in case someone would want to add more to it. Might not be a big issue since you initialize said button here, but note the use of `classList.add()` and `classList.remove()` in case you want to modify a class.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +64,135 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  })
+}
+
+/** Get the new list of stops locally */
+function getStopsList(){
+  const stopList = document.getElementById('stop-list');
+  if(stopList != null){
+    // clear list
+    stopList.innerText = ""; 
+  }
```
Leo2452, line 95: When would stopList be null? Similar to a comment I left below in the `script.js`, consider only deleting the stop you have to delete rather than clearing them all and loading `n-1` elements back from the server.

PR #17, opened 2020-08-03
Routepage Selection

j-vent:
JS file to add and delete places from the recommendations and stops/schedule panels.

### DIFF
--- a/byway/src/main/webapp/routepage.js
+++ b/byway/src/main/webapp/routepage.js
@@ -13,10 +13,9 @@
 // limitations under the License.
 
 
-// copies of recommendations and selected stops, stored as arrays
-// for synchronous updating
+// copies of recommendations and selected stops, stored as arrays for synchronous updating
 let recs = [];
-let stops = [];
+let stops = []; // TODO: Make set to avoid duplicates
 
 if (document.readyState === 'loading') {  // Loading hasn't finished yet
   document.addEventListener('DOMContentLoaded', loadData);
@@ -30,6 +29,7 @@ function loadData(){
   getStopsOnload();
 }
 
+/** Initializes map on the page */
 function initMap() {
   let directionsService = new google.maps.DirectionsService();
   let directionsRenderer = new google.maps.DirectionsRenderer();
@@ -45,7 +45,6 @@ function initMap() {
     calcRoute(directionsService, directionsRenderer, start, end);
   });
 }
-
 
 /** 
  * Displays route containing waypoints overtop the map.
@@ -76,67 +75,49 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
 function clearStops(){
   const stopList = document.getElementById('stop-list');
   if(stopList != null){
-    stopList.innerText = ""; // clear list
+    stopList.innerText = "";
   }
 }
 
 /** Get the new list of stops from datastore onload */
 function getStopsOnload(){
-  console.log("getstopsOnloda");
   clearStops();
   fetch('/api/stop')
   .then(response => response.json())
   .then((stopsResponse) => {
     stopsResponse.forEach((stop)=>{
-      console.log("pushed into stops " + stop);
       stops.push(stop);
     });
     renderStopsList();
   });
-  
 }
 
-/** Get the new list of stops locally */
+/** Re-render stop list synchronously */
 function renderStopsList(){
-  console.log("renderstoplist");
   clearStops();
-  // re-render list synchronously
   stops.forEach((stop)=>{
     renderStop(stop);
   })
 }
 
-/** Render stop list  */
+/** Render stop list  
+ *  @param {String} stop a String to add as a button in the schedule panel in the html
+ */
 function renderStop(stop){
-  console.log("render stop");
   const stopList = document.getElementById('stop-list');
-  // add to stop array in the js 
   let btn = document.createElement('button');
   btn.innerText = stop;
   btn.className =  "btn rec-btn";
   btn.addEventListener("click", function() {
-    // deleteFromStops(stop); 
     updateStops(stop, "remove");
   });
   stopList.appendChild(btn);
 }
 
-/** 
-/** Add stop locally and to datastore 
-function addToStops(stop){
-  deleteFromRecs(stop); 
-  // add to stops array locally in js
-  stops.push(stop);
-  renderStopsList();
-
-  // add to datastore
-  const params = new URLSearchParams();
-  params.append("text", stop.placename);
-  params.append("action", "add");
-  fetch('/api/stop', {method: 'POST', body: params});
-}
-*/
-
+/** Add stop to or delete stop from the stoplist in javascript and in the datastore
+ *  @param {String} stop a String to add or delete
+ *  @param {String} action a String to determine what action to perform on the list
+ */
 function updateStops(stop, action){
   if(action === "add"){
     // TODO: delete from recs later for better visuals on html
@@ -149,29 +130,11 @@ function updateStops(stop, action){
   }
   renderStopsList();
   let stopsAsJSONString = JSON.stringify(stops);
-  console.log("JSON:" + stopsAsJSONString);
   let params = new URLSearchParams();
   params.append("stops", stopsAsJSONString);
   fetch('/api/stop', {method: 'POST', body:params});
 }
 
-/** Delete stop locally and from datastore
-function deleteFromStops(stop){
-  addToRecs(stop); 
-  console.log("delete " + stop);
-  // delete from stops array locally in js
-  stops = stops.filter(function(stopObj){
-    return stopObj != stopsResponse;
-  })
-  renderStopsList();
-
-  // delete from datastore
-  const params = new URLSearchParams();
-  params.append("text", stop);
-  params.append("action", "remove");
-  fetch('/api/stop', {method: 'POST', body: params});
-}
-*/
 /** Clear the recommendations panel in the html */
 function clearRecs(){
   const recList = document.getElementById('rec-list');
@@ -180,63 +143,39 @@ function clearRecs(){
   }
 }
 
-
-
 /** Get the new list of recommendations from servlet onload */
 function getRecsOnload() {
-  console.log("getRecsOnloda");
   clearRecs();
   fetch('/api/recs')
   .then(response => response.json())
   .then((recommendations) => {
     recommendations.forEach((rec)=>{
       recs.push(rec);
-      console.log("pushed" + rec);
-      // renderRecs(rec);
     })
     renderRecsList();
   })
-  
 }
 
-/** Get the new list of recommendations locally */
+/** Re-render recs list synchronously */
 function renderRecsList(){
-  console.log("renderrecslist");
   clearRecs();
-  // re-render list synchronously
-  console.log(recs);
   recs.forEach((rec)=>{
-    // console.log("in foreach");
     renderRec(rec);
   })
 }
 
-/** Render recommendations list */
+/** Render stop list  
+ *  @param {String} rec a String to add as a button in the recommendations panel in the html
+ */
 function renderRec(rec){
-  console.log("renderrec");
   const recsList = document.getElementById('rec-list');
   let btn = document.createElement('button');
   btn.innerText = rec;
   btn.className =  "btn rec-btn";
   btn.addEventListener("click", function() {
-    // addToStops(rec); 
     updateStops(rec, "add");
   });
   recsList.appendChild(btn);
-}
-
-/** Add recommendation locally*/
-function addToRecs(rec){
-  recs.push(rec);
-  renderRecsList();
-}
-
-/** Delete recommendation locally */
-function deleteFromRecs(rec){
-  recs = recs.filter(function(r){
-    return r.placename != rec.placename;
-  })
-  renderRecsList();
 }
 
 /* exported initMap */
--- a/byway/src/main/webapp/routepage.js
+++ b/byway/src/main/webapp/routepage.js
@@ -80,21 +80,6 @@ function clearStops(){
   }
 }
 
-/** Render stop list  */
-function renderStop(stop){
-  const stopList = document.getElementById('stop-list');
-  // add to stop array in the js 
-  
-  let btn = document.createElement('button');
-  btn.id = stop.id;
-  btn.innerText = stop.placename;
-  btn.className =  "btn rec-btn";
-  btn.addEventListener("click", function() {
-    deleteFromStops(stop); 
-  });
-  stopList.appendChild(btn);
-}
-
 /** Get the new list of stops from datastore onload */
 function getStopsOnload(){
   console.log("getstopsOnloda");
@@ -103,8 +88,8 @@ function getStopsOnload(){
   .then(response => response.json())
   .then((stopsResponse) => {
     stopsResponse.forEach((stop)=>{
+      console.log("pushed into stops " + stop);
       stops.push(stop);
-      // renderStops(stop);
     });
     renderStopsList();
   });
@@ -121,7 +106,23 @@ function renderStopsList(){
   })
 }
 
-/** Add stop locally and to datastore */
+/** Render stop list  */
+function renderStop(stop){
+  console.log("render stop");
+  const stopList = document.getElementById('stop-list');
+  // add to stop array in the js 
+  let btn = document.createElement('button');
+  btn.innerText = stop;
+  btn.className =  "btn rec-btn";
+  btn.addEventListener("click", function() {
+    // deleteFromStops(stop); 
+    updateStops(stop, "remove");
+  });
+  stopList.appendChild(btn);
+}
+
+/** 
+/** Add stop locally and to datastore 
 function addToStops(stop){
   deleteFromRecs(stop); 
   // add to stops array locally in js
@@ -134,24 +135,43 @@ function addToStops(stop){
   params.append("action", "add");
   fetch('/api/stop', {method: 'POST', body: params});
 }
+*/
 
-/** Delete stop locally and from datastore*/
+function updateStops(stop, action){
+  if(action === "add"){
+    // TODO: delete from recs later for better visuals on html
+    stops.push(stop);
+  } else{
+    // TODO: add to recs later for better visuals on html
+    stops = stops.filter(function(stopObj){
+    return stopObj != stop;
+    })
+  }
+  renderStopsList();
+  let stopsAsJSONString = JSON.stringify(stops);
+  console.log("JSON:" + stopsAsJSONString);
+  let params = new URLSearchParams();
+  params.append("stops", stopsAsJSONString);
+  fetch('/api/stop', {method: 'POST', body:params});
+}
+
+/** Delete stop locally and from datastore
 function deleteFromStops(stop){
   addToRecs(stop); 
-  console.log("delete " + stop.placename);
+  console.log("delete " + stop);
   // delete from stops array locally in js
   stops = stops.filter(function(stopObj){
-    return stopObj.placename != stop.placename;
+    return stopObj != stopsResponse;
   })
   renderStopsList();
 
   // delete from datastore
   const params = new URLSearchParams();
-  params.append("text", stop.placename);
+  params.append("text", stop);
   params.append("action", "remove");
   fetch('/api/stop', {method: 'POST', body: params});
 }
-
+*/
 /** Clear the recommendations panel in the html */
 function clearRecs(){
   const recList = document.getElementById('rec-list');
@@ -160,19 +180,7 @@ function clearRecs(){
   }
 }
 
-/** Render recommendations list */
-function renderRec(rec){
-  console.log("renderrec");
-  const recsList = document.getElementById('rec-list');
-  let btn = document.createElement('button');
-  btn.id = rec.id;
-  btn.innerText = rec.placename;
-  btn.className =  "btn rec-btn";
-  btn.addEventListener("click", function() {
-    addToStops(rec); 
-  });
-  recsList.appendChild(btn);
-}
+
 
 /** Get the new list of recommendations from servlet onload */
 function getRecsOnload() {
@@ -183,7 +191,7 @@ function getRecsOnload() {
   .then((recommendations) => {
     recommendations.forEach((rec)=>{
       recs.push(rec);
-      console.log("pushed")
+      console.log("pushed" + rec);
       // renderRecs(rec);
     })
     renderRecsList();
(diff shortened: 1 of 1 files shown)


byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,139 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+
+  fetch('/api/stop')
+  .then(response => response.json())
+  .then((stopsResponse) => {
+    stopsResponse.forEach((stop)=>{
+      stops.push(stop);
+      var btn = document.createElement('button');
+      btn.id = stop.id;
+      btn.innerText = stop.placename;
+      btn.setAttribute("class", "btn rec-btn");
```
Leo2452, line 82: To remain consistent, consider accessing/setting the attribute by `btn.className = "btn rec-btn"` rather than using `setAttribute`

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,139 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+      btn.addEventListener("click", function() {
+        deleteFromStops(stop); 
+      });
+     stopList.appendChild(btn);
+    })
+  })
+}
+
+/** Get the new list of stops locally */
+function getStopsList(){
```
Leo2452, line 92: Not sure what the difference between `getStopsList()` and `getStops()` is. I see one gets from the server and other doesn't, but rather than clear them all and then add `n+1` stops, consider adding just the one stop by seeing if its already included in local array. If not, add the one new button, or else remove the one said button.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,139 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+
+  // re-render list synchronously
+  stops.forEach((stop)=>{
+    var btn = document.createElement('button');
+    btn.innerText = stop.placename
+    btn.setAttribute("class", "btn rec-btn");
+    btn.addEventListener("click", function() {
+      deleteFromStops(stop); 
+    });
+    stopList.appendChild(btn);
```
Leo2452, line 107: Can decompose this into a separate function to use across `getStops()` and `getStopsList()`

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,139 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+    recommendations.forEach((rec)=>{
+      // populate local js array initially
+      recs.push(rec);
+      var btn = document.createElement('button');
+      btn.id = rec.id;
+      btn.innerText = rec.placename;
       btn.setAttribute("class", "btn rec-btn");
       btn.addEventListener("click", function() {
-        addToStops(stop);
+        addToStops(rec);
```
Leo2452, line 160: Wouldn't you addToRecs instead, since they are recommendations and loading them at first?

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,139 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  const recsList = document.getElementById('rec-list');
+  recsList.innerText = ""; // clear list
+  // re-render list synchronously
+  recs.forEach((rec)=>{
+    var btn = document.createElement('button');
+    btn.innerText = rec.placename
+    btn.setAttribute("class", "btn rec-btn");
+    btn.addEventListener("click", function() {
+      addToStops(rec);
     });
```
Leo2452, line 178: Similarly, can decompose since this is a similar functionality to what you're doing in `getRecs()`

byway/src/main/webapp/routepage.js
j-vent: Oh no... in my master stuff I also had this format. I'll make a note to change that, but this is changed.

byway/src/main/webapp/routepage.js
j-vent: Yes, this was brought up as a suggestion from Landon. My only issue is that what happens to the placement if I remove the button, does everything get shifted? This way was just clearer for me to implement.

byway/src/main/webapp/routepage.js
j-vent: Ah yes, there's only like one line of difference. Will do, thanks for pointing that out!

byway/src/main/webapp/routepage.js
j-vent: I think you misunderstood, this is the function called upon clicking of a recommendation, which adds it to the stops panel. I have the recs.push(rec) above(w a comment) which adds to the list of recommendations.

byway/src/main/webapp/routepage.js
j-vent: Done!

byway/src/main/webapp/routepage.js
```diff
@@ -12,10 +12,22 @@
...
 if (document.readyState === 'loading') {  // Loading hasn't finished yet
-  document.addEventListener('DOMContentLoaded', getRecs);
+  document.addEventListener('DOMContentLoaded', load);
 } else {  // `DOMContentLoaded` has already fired
-  getRecs();
+  load();
+}
+
+/** Used to restore stops and recommendations upon load or refresh */
+function load(){
```
landondare, line 28: nit: prefer slightly more descriptive name, like `loadData`

byway/src/main/webapp/routepage.js
```diff
@@ -29,11 +41,13 @@ function initMap() {
...
   let map = new google.maps.Map(document.getElementById('map'), mapOptions);
   directionsRenderer.setMap(map);
-    document.getElementById("route").addEventListener("click", function() {
+  document.getElementById("route").addEventListener("click", function() {
     calcRoute(directionsService, directionsRenderer, start, end);
   });
 }
 
+
+/** Displays route overtop the map */
```
landondare, line 50: Please add documentation on what parameter this function and others expect.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  const stopList = document.getElementById('stop-list');
+  if(stopList != null){
+    stopList.innerText = ""; // clear list
+  }
 }
 
-function getRecs() {
-  fetch('/api/recs')
+/** Render stop list  */
+function renderStops(stop, fromDatastore){
```
landondare, line 75: Please use a more descriptive name for the parameters. For example, boolean parameters are often prefixed with `is-` to specify that it is a true or false value.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
-      let btn = document.createElement("button");
-      btn.innerText = stop.placename;
-      btn.setAttribute("class", "btn rec-btn");
-      btn.addEventListener("click", function() {
-        addToStops(stop);
-      });
-      recList.appendChild(btn);
+  .then((stopsResponse) => {
+    stopsResponse.forEach((stop)=>{
+      renderStops(stop, true);
```
landondare, line 98: Just add the `stop` to the list of stops here rather than in the `renderStops()` function. It is not really a part of the render logic anyways, and  you can get rid of the extra boolean param.

byway/src/main/webapp/routepage.js
landondare: If it is easier to replace all items, then let's just do that for the time being. It's not uncommon to redraw an entire UI element once the underlying data change, though it can be inefficient. We can always optimize this later.

byway/src/main/webapp/routepage.js
landondare, line 75: It would be more accurate to call this `renderStop()` since it only renders a single stop, not multiple.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
     });
+  })
+}
+
+/** Get the new list of stops locally */
+function getStopsList(){
```
landondare, line 104: This function name is a bit inaccurate. You are not actually getting a list of stops, rather you are clearing and redrawing a list of stops you already have. It would be more accurate to call this something like `renderStopList()`

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  clearStops();
+  // re-render list synchronously
+  stops.forEach((stop)=>{
+    renderStops(stop, false);
+  })
+}
+
+/** Add stop locally and to datastore */
+function addToStops(stop){
```
landondare, line 113: Instead of having an add and remove function, it would be better to have a single `updateStops()` method. Here, you would simply pass the entire list  of stops rather than the single item you want to add or remove. This objects are simplistic enough that you can can pass the whole list to the server

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  deleteFromRecs(stop); 
+  // add to stops array locally in js
+  stops.push(stop);
+  getStopsList();
+
+  // add to datastore
+  const params = new URLSearchParams();
+  params.append("text", stop.placename);
+  params.append("action", "add");
```
landondare, line 122: Quick note: Encoding logic like this through a url parameter is not a good design pattern. It would be better to have a separate endpoint, or to make a different type of request. 

A pattern you will see more common is instead of specifying how the data is changing, you just pass the whole updated object. E.g. you remove or add what you want, and the server just replaces the whole object with whatever you provide.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+/** Clear the recommendations panel in the html */
+function clearRecs(){
+  const recList = document.getElementById('rec-list');
+  if(recList != null){
+    recList.innerText = ""; // clear list
+  }
+}
+
+/** Render recommendations list */
+function renderRecs(rec, fromDatastore){
```
landondare, line 152: Same comment as above, use singular `Rec` if this only acts on a single recommendation.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  const recsList = document.getElementById('rec-list');
+  // populate local js array initially
+  if(fromDatastore){
```
landondare, line 155: Same comment as before, add the recommendation to this list when you get it from data store, rather than when you are rendering it.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  recsList.appendChild(btn);
+}
+
+/** Get the new list of recommendations from servlet onload */
+function getRecsOnload() {
+  clearRecs();
+  fetch('/api/recs')
+  .then(response => response.json())
+  .then((recommendations) => {
+    recommendations.forEach((rec)=>{
```
landondare, line 174: You can just add all the recommendations to your local variable here and then call the `getRecsList()` functions. It performs basically the same logic. Generally it is a good practice to centralize repeated logic where you can. A good strategy to follow is DRY: Don't Repeat Yourself.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +63,145 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  recs.forEach((rec)=>{
+      renderRecs(rec, false);
+  })
+}
+
+/** Add recommendation locally*/
+function addToRecs(rec){
+  recs.push(rec);
+  getRecsList();
+  const params = new URLSearchParams();
```
landondare, line 193: What is this used for? Did you mean to delete these two lines?

byway/src/main/webapp/routepage.js
j-vent: Done!

byway/src/main/webapp/routepage.js
j-vent: done!

byway/src/main/webapp/routepage.js
j-vent: Okay, I know about `isCondition` but I thought `isFromDatastore` looked clunky. Will change back.

byway/src/main/webapp/routepage.js
j-vent: Oh right, that was an unnecessary complication, thanks for pointing that out.

byway/src/main/webapp/routepage.js
j-vent: Okay, sounds good.

byway/src/main/webapp/routepage.js
j-vent: Done!

byway/src/main/webapp/routepage.js
j-vent: Yup, that's true. Changed!

byway/src/main/webapp/routepage.js
j-vent: done!

byway/src/main/webapp/routepage.js
j-vent: done!

byway/src/main/webapp/routepage.js
j-vent: Oh yes, that's a mistake. I originally was changing the recommendations in the datastore, but you suggested not removing anything from that. Right now, if you refresh the page you get the same hardcoded places rendered as there is no deletion/addition. but the schedule/selected stops panel does get updated properly. 

Also there's a lot of uncertainty with getting the recommendations into datastore that I will address in a future pull request.

byway/src/main/webapp/routepage.js
j-vent: Yeah... it looked janky but it worked. I will switch to your suggested method of updating/sending the entire list instead as that mitigates this issue.

byway/src/main/webapp/routepage.js
j-vent: Good point, thanks for pointing that out.

byway/src/main/webapp/routepage.js
j-vent: I used a simplified Trip class to simulate storing and getting destinations in this updateStops() method. I also made it so you can either add or remove a place all in the same function, but I could split it up for readability. Please let me know what is best.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,113 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+/** Render stop list  
+ *  @param {String} stop a String to add as a button in the schedule panel in the html
+ */
+function renderStop(stop){
+  const stopList = document.getElementById('stop-list');
+  let btn = document.createElement('button');
+  btn.innerText = stop;
+  btn.className =  "btn rec-btn";
+  btn.addEventListener("click", function() {
+    updateStops(stop, "remove");
```
landondare, line 112: You can remove the stop from the `stops` list and then call update. This way you don't need to pass the action you want to perform, you can do it directly here.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,113 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+/** Render stop list  
+ *  @param {String} rec a String to add as a button in the recommendations panel in the html
+ */
+function renderRec(rec){
+  const recsList = document.getElementById('rec-list');
+  let btn = document.createElement('button');
+  btn.innerText = rec;
+  btn.className =  "btn rec-btn";
+  btn.addEventListener("click", function() {
+    updateStops(rec, "add");
```
landondare, line 176: Same here, you could add the rec to the list here.

byway/src/main/webapp/routepage.js
j-vent: Right, I have a tendency to add unnecessary params. Changed!

byway/src/main/webapp/routepage.js
j-vent: Done!

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,110 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+ */
+function renderRec(rec){
+  const recsList = document.getElementById('rec-list');
+  let btn = document.createElement('button');
+  btn.innerText = rec;
+  btn.className =  "btn rec-btn";
+  btn.addEventListener("click", function() {
+    // TODO: add to recs later for better visuals on html
+    stops.push(rec);
+    updateStops(rec);
```
zstewar1, line 173: updateStops appears to take no arguments where it is defined above.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,110 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  }
+}
+
+/** Get the new list of recommendations from servlet onload */
+function getRecsOnload() {
+  clearRecs();
+  fetch('/api/recs')
+  .then(response => response.json())
+  .then((recommendations) => {
+    recommendations.forEach((rec)=>{
```
zstewar1, line 147: Note that since we're already using modern JS features, you can use the "spread" operator to push all recommendations at once, rather than using forEach.

`recs.push(...recommendations);`

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,110 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  recs.forEach((rec)=>{
+    renderRec(rec);
+  })
+}
+
+/** Render stop list  
+ *  @param {String} rec a String to add as a button in the recommendations panel in the html
+ */
+function renderRec(rec){
+  const recsList = document.getElementById('rec-list');
```
zstewar1, line 166: Consider separating creating the buttons from deciding where to insert them into the DOM. E.g. you can make this function be `createRecButton`, and just return the `btn` element. Then the calling code would decide where to do the insertions, like:

```js
function renderRecsList(){
  clearRecs();
  const recsList = document.getElementById('rec-list');
  recs.forEach((rec)=>{
    recsList.appendChild(createRecButton(rec));
  })
}
```

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,110 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+
+/** Add stop to or delete stop from the stoplist in javascript and in the datastore
+ *  @param {String} stop a String to add or delete
+ */
+function updateStops(){
+  renderStopsList();
+  let stopsAsJSONString = JSON.stringify(stops);
+  let params = new URLSearchParams();
+  params.append("stops", stopsAsJSONString);
+  fetch('/api/stop', {method: 'POST', body:params});
```
zstewar1, line 129: When sending json data in the body, it should be sufficient to just put `body: JSON.stringify(stops)`. You can ask Leo about how we retrieve this a JSON request body, since that was covered in #18

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,110 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+  clearStops();
+  stops.forEach((stop)=>{
+    renderStop(stop);
+  })
+}
+
+/** Render stop list  
+ *  @param {String} stop a String to add as a button in the schedule panel in the html
+ */
+function renderStop(stop){
```
zstewar1, line 106: Same advice as below for using a `createStopButton` and letting the caller decide where it should go on the page.

byway/src/main/webapp/routepage.js
j-vent: Yup, noticed this on another branch, but did not change it back here. Thanks!

byway/src/main/webapp/routepage.js
j-vent: Okay, thanks for suggesting a cleaner design that's understood more easily. Changed.

byway/src/main/webapp/routepage.js
j-vent: Done!

byway/src/main/webapp/routepage.js
j-vent: Okay, thank you for the suggestion. I've changed all occurrences of forEach().

byway/src/main/webapp/routepage.js
j-vent: Okay, will remove the extra variables. I also changed the servlet code to resemble what Leo had, but I'll read through your exchange in more depth.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,105 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+    stopList.appendChild(createStopButton(stop));
+  })
+}
+
+/** Render stop list  
+ *  @param {String} stop a String to add as a button in the schedule panel in the html
+ *  @return {button} stopBtn a button showing a selected stop
+ */
+function createStopButton(stop){
+  let stopBtn = document.createElement('button');
```
zstewar1, line 109: Note that `const` in JS means "cannot assign a new value to the variable", it doesn't mean "can't modify the variable".

```js
const example = [];
example.push(3); // this is fine because it just modifies the value, doesn't assign to the variable.
example = [3]; // this doesn't work because it tries to assign a new value to the variable.
```

Aside from preventing re-assignment, `const` and `let` are identical, so you should always use `const` unless you specifically need to be able to assign a new value to the variable. Using the stricter variable type helps prevent bugs and makes the program easier to read and understand, because you know when you see `const` that the variable won't be reassigned anywhere, so it's a little easier to reason about. This is the essentially the same reason we always say to use `final` on fields of classes when working in Java.

byway/src/main/webapp/routepage.js
```diff
@@ -49,27 +71,105 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+     recsList.appendChild(createRecButton(rec));
+  })
+}
+
+/** Render stop list  
+ *  @param {String} rec a String to add as a button in the recommendations panel in the html
+ *  @return {button} recBtn a button showing a recommended place
+ */
+function createRecButton(rec){
+  let recBtn = document.createElement('button');
```
zstewar1, line 162: `const`, same as above.

byway/src/main/webapp/routepage.js
j-vent: Oh I see, I thought it meant can't modify so I have been using let this whole time. Will change this and other occurences.

byway/src/main/webapp/routepage.js
j-vent: done!

PR #18, opened 2020-08-03
Store interests from retrieved user

Leo2452:
When users select all interests, it sends a form to the servlet to search for the current user (with hard-coded, temporary ID) and adds a property to store their interests inside their specific entity type.


byway/src/main/webapp/track-interests.js
```diff
@@ -13,9 +13,33 @@
...
+ * Sets properties to the interest form on the html
+ * page. Handles submit event to keep user on the
+ * same page and send information.
+ */
+function loadForm() {
+  let interestsForm = document.getElementById('interests-form');
+  interestsForm.addEventListener('submit', () => {
+    let interestsAsJSONString = JSON.stringify(Array.from(interestsChosen));
+    let params = new URLSearchParams();
+    params.append("data", interestsAsJSONString);
```
j-vent, line 40: Change "data" to interests so it's more specific.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,39 @@ private String formatLocation(PlaceType location) {
...
+    Query person =
+      new Query("user")
+        .setFilter(new FilterPredicate("id", FilterOperator.EQUAL, 2452));
+    PreparedQuery pq = datastore.prepare(person);
+    Entity userEntity = pq.asSingleEntity();
+    if(userEntity != null) {
+      String interestsAsString = request.getParameter("data");
+      addInterestsForUser(interestsAsString, userEntity, datastore);
+    } else {
+      response.setStatus(response.SC_NOT_FOUND);
```
j-vent, line 79: I don't think you should hardcode the status, because you can't guarantee that it will be a 404 error.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,39 @@ private String formatLocation(PlaceType location) {
     place = place.replace('_', ' ');
     return place;
   }
+
+  /**
+   * Temporary setup to find entity with hard-coded id value. Set their interests.
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
j-vent, line 69: Just have datastore as a private final variable instead of passing it as a param.

byway/src/main/webapp/track-interests.js
```diff
@@ -13,9 +13,33 @@
...
+    fetch('/api/places', {method: 'POST', body: params});
+  })
 }
 
 let interestsChosen = new Set();
```
j-vent, line 45: nit. Line 86. Can you change the className button to something more specific? I'm using the btn class from bootstrap, which is a reserved keyword.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +60,39 @@ private String formatLocation(PlaceType location) {
...
+
+
+  /**
+   * Creates a list of interests from the JSON string passed in and adds it
+   * to this user's Entity object as a separate property.
+   * @param interestsAsString JSON String containing interests.
+   * @param currentUser       Entity with user information.
+   * @param datastore         DatastoreService with all users.
+   */
+  private void addInterestsForUser(String interestsAsString, Entity currentUser,
```
renau06, line 91: if you declare datastore before your doPost you wont need to pass it as a param

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: I set the status because it would indicate a user wasn't found in datastore. What do you recommend I do to handle this case? I set it as a not found error, albeit not the same thing.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Fixed

byway/src/main/webapp/track-interests.js
Leo2452: Will do, fixed

byway/src/main/webapp/track-interests.js
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: I added a comment to show the entity must exist and cannot be null since it's made beforehand, and referenced with a specific unique trip id to guarantee it's found and a single entity. Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +64,35 @@ private String formatLocation(PlaceType location) {
...
+   * @param interestsAsString JSON String containing interests.
+   * @param currentUser       Entity with user information.
+   * @param datastore         DatastoreService with all users.
+   */
+  private void addInterestsForTrip(String interestsAsString, Entity trip) throws IOException {
+
+    ArrayList<String> interests = gson.fromJson(interestsAsString, ArrayList.class);
+    trip.setProperty("interests", interests);
+    datastore.put(trip);
+  }
```
zstewar1, line 97: I'd try to break this down into two operations a little differently. Right now you have the caller get the parameter from the request, then this function combines the operations  of (parse json string, apply update to entity). To me, this doesn't seem like a very logical grouping, since it's very specific to this use case:

caller: (Get json string, Get entity)
function: (Parse json string, update entity)

Notice how this grouping has both the caller and the called function operating with both json strings and entities. Additionally, the caller doesn't have any flexibility regarding how they use the parsed data.

Consider this alternate grouping of functionality:

function 1: (Get json string, Parse json string)
function 2: (Get entity, update entity)

Each function operates on only a single type of data, either json or entities, but *not* both, with the intermediary between them being `ArrayList<String>`. this separation of which kind of data the functions operate on makes it much easier to generalize them in the future, if you find yourself doing similar work in many places.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +64,35 @@ private String formatLocation(PlaceType location) {
...
+  /**
+   * Creates a list of interests from the JSON string passed in and adds it
+   * to this user's Entity object as a separate property.
+   * @param interestsAsString JSON String containing interests.
+   * @param currentUser       Entity with user information.
+   * @param datastore         DatastoreService with all users.
+   */
+  private void addInterestsForTrip(String interestsAsString, Entity trip) throws IOException {
+
+    ArrayList<String> interests = gson.fromJson(interestsAsString, ArrayList.class);
```
zstewar1, line 94: Note that when you use `fromJson(..., ArrayList.class)`, gson has no way of knowing what type the contents of the array list should be and has to guess. This could result in unexpected behavior, for example if the calling code includes an integer in the input by mistake, you might see a `ClassCastException` much later on when you are trying to access the contents of the `ArrayList`, or possibly even more confusingly, after successfully storing the list in datastore, and pulling it back out, you could potentially see an error when reading values from the ArrayList from datastore.

This is becuase Java's generics only exist at compile time. In actual fact, when running the program, `ArrayList<String>` doesn't exist, there is only plain `ArrayList`, which is basically like writing `ArrayList<Object>`. This means that even though this says `interests` is `ArrayList<String>`, it could very possibly contain objects of some other type like `Integer`. (You can try this out yourself by creating an `ArrayList<Integer>`, then casting it to `ArrayList<Object>` then casting that to `ArrayList<String>`. This will (should) compile, and it will allow you to store the wrong types of objects in the list, since types are not checked when adding to the list, but it will cause a `ClassCastException` if you try to take something out of the list that's the wrong type.

Essentially, the Java compiler first does it's best to ensure that generic types you have written has the correct types, then it does what is called 'type erasure', where it essentially removes all the generics and inserts a bunch of casts whenever you access something generic.

So if you write an obviously wrong like:

```java
ArrayList<String> l1 = new ArrayList<>();
l1.add("hello world");
ArrayList<Integer> l2 = l1; // fails at compile time because the generic types are different.
```

the compiler will catch it. But the compiler will respect when you cast types, so you can write something wrong like this:

```java
ArrayList<Integer> intList = new ArrayList<>();
ArrayList<Object> objList = (ArrayList<Object>) intList;
ArrayList<String> strList = (ArrayList<String>) objList;
intList.add(1);
strList.add("Hello world");

int intFromList = intList.get(0); // this works because intList[0] is actually an Integer.
String strFromList = strList.get(1); // this works because strList[1] is actually a String.
int intFromList = intList.get(1); // this throws ClassCastException because intList[1] is "Hello world"
```

it won't stop you and you'll get an error at runtime. The reason it works this way is because, after being compiled, the above code is equivalent to this:

```java
ArrayList intList = new ArrayList();
ArrayList objList = intList;
ArrayList strList = objList;
// Notice how all three lists reference the same list now.
intList.add(1);
strList.add("Hello world");

int intFromList = (Integer) intList.get(0);
String strFromList = (String) strList.get(1);
int intFromList = (Integer) intList.get(1);
```

Basically, the generics are dropped, and the running program works as if everything is an Object, and the compiler inserts casts for you wherever you access the contents of the generic type, in this case through `ArrayList.get(int index)`. And that's why if your list contains the wrong type, you might not find out unless you try to access the contents.

-----

## So how to avoid this when deserializing?

You want to tell gson what actual type you expect, so it will only deserialize the correct generic type, so you can't accidentally get an `ArrayList<String>` containing integers or whatever to cause problems later on.

You can do this by using a [`TypeToken`](https://www.javadoc.io/doc/com.google.code.gson/gson/2.6.2/com/google/gson/reflect/TypeToken.html).

First import `Type` and `TypeToken`:

```java
import com.google.gson.reflect.TypeToken;
import java.lang.reflect.Type;
```

Then, create a constant for the type you are trying to parse:

```java
/** {@link Type} of an {@link ArrayList} containing {@link String}, for gson decoding. */
private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
```

then just switch `ArrayList.class` in the `gson.fromJson` for the `ARRAYLIST_STRING` constant.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +64,35 @@ private String formatLocation(PlaceType location) {
...
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query specificTrip =
+      new Query("trip")
+        .setFilter(new FilterPredicate("id", FilterOperator.EQUAL, 2452));
+    PreparedQuery pq = datastore.prepare(specificTrip);
+
+    // Guaranteed to have one entity, since id would be unique.
+    Entity trip = pq.asSingleEntity();
```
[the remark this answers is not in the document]
Leo2452: Good point, I was unsure how to convert from a key's id back to it's key. I forgot to think about the entity `kind`, and found that `KeyFactory` can create the key using these two params. Thanks, fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: I agree this setup is somewhat odd/very specific. Fixed it to separate the tasks more reasonably

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Very cool and insightful! I didn't consider to use the type rather than the class for `fromJson()`, partly because I'm not too sure of the difference and I'm not sure how Java handles things in compile when it deals with/compiles generic types. I'll look more into this type erasure stuff and probably ask for some clarification if I stumble or to affirm if I understand things well.

Thanks, and fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +71,44 @@ private String formatLocation(PlaceType location) {
     place = place.replace('_', ' ');
     return place;
   }
+
+  /**
+   * Temporary setup to find entity with hard-coded id value. Set their interests.
```
landondare, line 76: This comments seems slightly out of date if I am not mistaken. Seems you are using the tripId provided in the request rather than a hard coded one.

Leo2452: Updated

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,45 @@
...
+ * Sets properties to the interest form on the html
+ * page. Handles submit event to keep user on the
+ * same page and send information.
+ */
+function loadForm() {
+  let interestsForm = document.getElementById('interests-form');
+  interestsForm.addEventListener('submit', (event) => {
+    event.preventDefault();
+    let interestsAsJSONString = JSON.stringify(Array.from(interestsChosen));
+    params.append("interests", interestsAsJSONString);
```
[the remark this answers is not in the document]
Leo2452: Fixed with FormData

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,45 @@
...
+  loadId();
   loadButtons();
+  loadForm();
 }
 
-let interestsChosen = new Set();
+/* Parse url to retrieve the trip id. */
+function loadId() {
+  const url = new URL(location.href);
+  params = new URLSearchParams(url.search);
```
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -14,21 +14,42 @@
...
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that returns all PlaceTypes from Places API. */
+
+/**
+ * Servlet that returns all PlaceTypes from Places API.
+ * Stores which placeTypes, or interests, user selects
+ * in datastore with their specific Trip Entity.
+ */
+@MultipartConfig
```
zstewar1, line 45: Note: I think you only use MultipartConfig when POST requests will be using FormData. If just sending json data, I don't think you need this.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +73,31 @@ private String formatLocation(PlaceType location) {
...
+    }
+  }
+
+  /**
+   * Converts the JSON of interests selected and parses it into an ArrayList of Strings.
+   * @param  request servlet request
+   * @return list of interests
+   */
+  private ArrayList<String> findInterests(HttpServletRequest request) {
+    String interestsAsJSON = request.getParameter("interests");
```
zstewar1, line 99: If you want to load something from the body of a post request rather than using query parameters, you can use `gson.fromJson(request.getReader(), ARRAYLIST_STRING)`. When doing this, the client would put the json string in the `body` of the post request, rather than adding it as query parameters.

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,48 @@
...
   loadButtons();
+  loadForm();
 }
 
-let interestsChosen = new Set();
+/* Parse url to retrieve the trip id. */
+function loadId() {
+  let queryString = location.search.substring(1);
+  let separatedData = queryString.split("&");
+  tripQuery = separatedData[0];
```
zstewar1, line 38: There are functions available for parsing query parameters, namely `URLSearchParams`. Don't try to split up data like that yourself.

Remember that URLs are user-visible, and anyone can put whatever they want in them. What happens if you try to split like this and there are query parameters that you didn't expect? What if you add more parameters in the future? Remember order in query parameters generally doesn't matter, so grabbing by index is a bad idea.

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,48 @@
...
+ */
+function loadForm() {
+  let interestsForm = document.getElementById('interests-form');
+  interestsForm.addEventListener('submit', (event) => {
+    event.preventDefault();
+    let interestsAsJSONString = JSON.stringify(Array.from(interestsChosen));
+    let formData = new FormData();
+    formData.append("interests", interestsAsJSONString);
+    fetch('/api/places?' + tripQuery,
+      {method: 'POST', body: formData});
```
zstewar1, line 54: Rather than writing out the logic for these kinds of fetches all over the place, consider making a common function that handles fetch with the correct path and query parameters. This will be especially useful if you start using a tripId for `GET` requests too.

For example,

```js
/**
 * Gets or updates the places on the server for the specified tripId. If no newInterests are specified, 
 * makes a get request to fetch the current places for the specified trip. Otherwise makes a post request
 * to update the places for the trip to the value specified. newInterests may be any type which can be
 * converted to an array by Array.from. It will be sent to the server as json in the post body. 
 */
function fetchPlaces(tripId, /* optional */ newInterests) {
  const url = '/api/places?' + new URLSearchParams({tripId}).toString();
  const fetchArgs = {method: 'GET'};
  if (newInterests !== undefined) {
    fetchArgs.method = 'POST';
    fetchArgs.body = JSON.stringify(Array.from(newInterests));
  }
  return fetch(url, fetchArgs).then(response => response.json());
}
```

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -14,21 +14,40 @@
...
+ * Stores which placeTypes, or interests, user selects
+ * in datastore with their specific Trip Entity.
+ */
 @WebServlet("/api/places")
 public final class PlacesServlet extends HttpServlet {
 
   private final Gson gson = new Gson();
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  /** {@link Type} of an {@link ArrayList} containing {@link String}, for gson decoding. */
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +71,44 @@ private String formatLocation(PlaceType location) {
...
     return place;
   }
+
+  /**
+   * Temporary setup to find entity with hard-coded id value. Set their interests.
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String tripIdString = request.getParameter("tripId");
+    Long tripId = Long.parseLong(tripIdString);
```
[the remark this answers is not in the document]
Leo2452: I actually changed this entirely since we are passing in the key as a string directly, rather than the id to said key. This avoids converting to id, then creating a new key. Instead, we pass the key as string, then convert that string into the key directly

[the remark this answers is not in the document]
Leo2452: Removed this approach, as mentioned above

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,45 @@
...
+function loadContent() {
+  loadId();
   loadButtons();
+  loadForm();
 }
 
-let interestsChosen = new Set();
+/* Parse url to retrieve the trip id. */
+function loadId() {
+  const url = new URL(location.href);
```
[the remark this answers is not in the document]

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Removed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java

byway/src/main/webapp/track-interests.js
Leo2452: Good point, fixed

byway/src/main/webapp/track-interests.js
Leo2452: Updated

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +72,21 @@ private String formatLocation(PlaceType location) {
...
+    String tripKeyString = request.getParameter("tripId");
+    Key tripKey = KeyFactory.stringToKey(tripKeyString);
+    ArrayList<String> interests = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    Entity trip;
+    try {
+      trip = datastore.get(tripKey);
+      trip.setProperty("interests", interests);
+      datastore.put(trip);
+      response.setContentType("application/json;");
+      response.getWriter().println("{\"results\": \"success\"}");
```
landondare, line 87: Do you need to return this "success" value here? The frontend should know whether the operation was successful based on the return code of the response. Unless you're returning something meaningful, you probably don't need to this.

byway/src/main/webapp/track-interests.js
```diff
@@ -59,6 +88,22 @@ function createButtonForPlace(place) {
...
+  button.className = "interestBtn";
   return button;
 }
+
+/**
+ * Gets all potential interests from server to load or sets the user's
+ * specific interests for their specific trip with tripQuery. userInterests
+ * may be any type that is convertible to an array via Array.from. It will
+ * be sent to the server as JSON in the post body.
+ */
```
landondare, line 100: Nit: add parameter documentation.

Leo2452: Updated

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Updated

byway/src/main/webapp/track-interests.js
```diff
@@ -59,6 +88,25 @@ function createButtonForPlace(place) {
...
+/**
+ * Gets all potential interests from server to load or sets the user's
+ * specific interests for their specific trip with tripQuery. userInterests
+ * may be any type that is convertible to an array via Array.from. It will
+ * be sent to the server as JSON in the post body.
+ * @param {String tripQuery} value for tripId
+ * @param {Array} [userInterests] interests selected by user
+ */
+function fetchPlaces(tripQuery, /* optional */ userInterests) {
+  const url = '/api/places?' + tripQuery;
```
zstewar1, line 104: Prefer to use `URLSearchParams` to construct values for query parameters.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +72,19 @@ private String formatLocation(PlaceType location) {
...
   }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String tripKeyString = request.getParameter("tripId");
+    Key tripKey = KeyFactory.stringToKey(tripKeyString);
+    ArrayList<String> interests = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    Entity trip;
+    try {
+      trip = datastore.get(tripKey);
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +72,19 @@ private String formatLocation(PlaceType location) {
...
+      trip.setProperty("interests", interests);
+      datastore.put(trip);
+    } catch(EntityNotFoundException e) {
+      //thrown automatically
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,43 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+let interestsChosen = new Set();
+let tripQuery;
```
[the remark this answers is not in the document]

byway/src/main/webapp/track-interests.js

byway/src/main/webapp/interests.html
```diff
@@ -9,6 +9,10 @@
   <body>
     <div id="header">Some content</div>
     <p> Scroll through and select some interests</p>
-    <div id="interests"></div>
+    <div id="interests-section"></div>
+    <form id="interests-form">
+      <input type="hidden" id="interests" name="interests" />
+      <input type="submit" value="Next page" />
+    </form>
```
zstewar1, line 16: What is the purpose of this empty form for the next button? You can just add `addEventListener('click', ...)` to a button rather than create a form for it, if you aren't going to use the form data for anything.

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,43 @@
...
+ */
+function loadContent() {
+  loadId();
   loadButtons();
+  loadForm();
 }
 
-let interestsChosen = new Set();
+/* Parse url to retrieve the trip id. */
+function loadId() {
```
zstewar1, line 35: Try to be more descriptive: `getTripIdFromUrl`.

byway/src/main/webapp/track-interests.js
```diff
@@ -12,13 +12,43 @@
...
+  let params = new URLSearchParams(location.search);
+  tripId = params.get('tripId');
+}
+
+/**
+ * Sets properties to the interest form on the html
+ * page. Handles submit event to keep user on the
+ * same page and send information.
+ */
+function loadForm() {
```
zstewar1, line 45: `configureNextPageButton` or something more descriptive.

byway/src/main/webapp/track-interests.js
```diff
@@ -59,6 +88,25 @@ function createButtonForPlace(place) {
...
+ * Gets all potential interests from server to load or sets the user's
+ * specific interests for their specific trip with tripId. userInterests
+ * may be any type that is convertible to an array via Array.from. It will
+ * be sent to the server as JSON in the post body.
+ * @param {String tripId} value for tripId
+ * @param {Array} [userInterests] interests selected by user
+ */
+function fetchPlaces(tripId, /* optional */ userInterests) {
+  const url = '/api/places?' + new URLSearchParams({tripId}).toString();
+  const fetchArgs = {method: 'GET'};
```
zstewar1, line 105: Note that since you're now returning the `fetch` call in two separate places, you can just in-line the value of `fetchArgs` like, `fetch(url)` for the `GET` case (get is the default, so you don't have to set it), or `fetch(url, {method: POST, body: ...})` for the `POST` case.

byway/src/main/webapp/interests.html
Leo2452: I updated to make the form send the data and add another explicit link to the next page. I thought I could set the action to this form to link to the next page since it was working with an earlier implementation, but it does not do what I intend to do. Also, the action attribute isn't meant for that but rather to send the form data to that action page.

Fixed

byway/src/main/webapp/track-interests.js
Leo2452: Updated

byway/src/main/webapp/track-interests.js
Leo2452: Updated

byway/src/main/webapp/track-interests.js

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +70,22 @@ private String formatLocation(PlaceType location) {
...
+    Entity trip;
+    try {
+      trip = datastore.get(tripKey);
+    } catch(EntityNotFoundException e) {
+      //thrown automatically
+      logger.atInfo().withCause(e).log("Unable to find Trip Entity with key string %s", tripKeyString);
+      response.setStatus(response.SC_NOT_FOUND);
+      return;
+    }
+    trip.setProperty("interests", interests);
```
zstewar1, line 88: It would be good to avoid working with the string property names outside the `Trip` class -- the point of having a wrapper class like that is to encapsulate the knowledge of what these string property names are, so that using them is less error prone. If you typo `setProperty("interessts", interests)` you'll wonder why in some places you can set the interests fine, and in others it doesn't work. If you typo `trip.setInteressts(interests)`, you'll get a compiler error, which is much easier to diagnose. In general string property access should be kept to as few places as possible, and should be encapsulated within the wrapper types. Have a look at rena's PR for some suggestions I gave on how to use the wrapper lasses for data manipulation like this (and coordinate among yourselves on who will implement them).

byway/src/main/webapp/track-interests.js
```diff
@@ -26,23 +26,31 @@ if(document.readyState === 'loading') {
...
 }
 
 /**
  * Sets properties to the interest form on the html
  * page. Handles submit event to keep user on the
- * same page and send information.
+ * same page and send information through a fetch request.
  */
-function loadForm() {
+function loadFormToSaveInterests() {
```
zstewar1, line 53: `load` implies the function itself loads data from the server, but this just configures what is essentially the "save" button.

byway/src/main/webapp/track-interests.js
```diff
@@ -26,23 +26,31 @@ if(document.readyState === 'loading') {
...
   let interestsForm = document.getElementById('interests-form');
   interestsForm.addEventListener('submit', (event) => {
```
zstewar1, line 55: You're not actually using the data *from the form* here, so I don't know why you have it as a form. If you just want a button, you can just make a button and bind to the 'click' event instead of 'submit'. Forms are containers for (usually user-entered) data, so if you aren't using any data from the form, you don't need a form.

byway/src/main/webapp/track-interests.js
```diff
@@ -12,17 +12,57 @@
...
+  getTripIdFromUrl();
+  loadButtonsWithInterests();
+  loadFormToSaveInterests();
+}
+
+/* Parse url to retrieve the trip id and load the next page with it. */
+function getTripIdFromUrl() {
+  let params = new URLSearchParams(location.search);
+  tripId = params.get('tripId');
+  if(tripId === null) {
```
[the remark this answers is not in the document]
Leo2452: Fixed. I also created a `setTripIdFor()` function to load the trip id for the next page. I added specifications that the document requires an element with a specific id/href attribute, ie an <a> tag, for use across pages (specifically Rena's and mine).

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Updated with the Trip class

byway/src/main/webapp/track-interests.js
Leo2452: Removed the form

byway/src/main/webapp/track-interests.js

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -64,6 +66,15 @@ public Trip(
     this.route = new ArrayList<String>(route);
   }
 
+  /**
+   * Set the interests for the trip as plain text with the Collection passed in.
+   *
+   * @param interests collection of Strings indicating a user interest
+   */
+  public void setInterests(Collection<String> interests) {
+    this.interests = new ArrayList<>(interests);
```
zstewar1, line 75: No need to make `interests` non-final and allocate a new array, you can just use `interests.clear(); interests.addAll(interests);` One nice thing about this is that if the inner array of `interests` is large enough to hold the new number of interests, it avoids reallocation, and can just copy the values from the input.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -52,4 +74,13 @@ private String formatLocation(PlaceType location) {
     place = place.replace('_', ' ');
     return place;
   }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String tripKeyString = request.getParameter("tripId");
+    ArrayList<String> interests = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    Trip trip = Trip.getTrip(datastore, tripKeyString);
```
zstewar1, line 82: Might be good to check that the trip is not null, and set a `NOT_FOUND` status if it is missing.

byway/pom.xml
```diff
@@ -24,7 +24,7 @@
       <version>4.0.1</version>
       <scope>provided</scope>
     </dependency>
-   
+
     <dependency>
       <groupId>com.google.maps</groupId>
       <artifactId>google-maps-services</artifactId>
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Updated

byway/src/main/webapp/base.js
```diff
@@ -11,3 +11,29 @@ function setProgressBar(pageNumber){
...
+
+/**
+ * Sets the trip Id passed in to the next page, specified
+ * by the filename. If the trip Id is null, redirects the
+ * user to the page where they can make a new trip and have
+ * a valid trip Id to reference.
+ * Note: Requires an element in the document with id "next-button"
+ * with an href attribute, e.g. an 'a' tag.
+ */
+function setTripIdFor(tripId, filename) {
```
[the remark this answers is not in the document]

byway/src/main/java/com/google/sps/data/UserInfo.java
```diff
@@ -28,7 +28,7 @@
 /** A class to make a UserInfo type, containing an email, id and a list of trip IDs per user. */
 public final class UserInfo {
 
-  public static final String DATASTORE_ENTITY_KIND = "user";
+  public static final String DATASTORE_ENTITY_KIND = "userInfo";
```
zstewar1, line 31: Is this a change you agreed to as a group? Renaming an entity kind means that any old entities already in datastore still have the old kind, so while this kind of change is safe early in development, it does have consequences if you do it later on when you already have entities with this kind. There's no hard requirement that the kind string exactly match the class name, so it would be fine to leave this as `"user"`.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -14,20 +14,43 @@
...
+ * Servlet that returns all PlaceTypes from Places API. Stores which placeTypes, or interests, user
+ * selects in datastore with their specific Trip Entity.
+ */
 @WebServlet("/api/places")
 public final class PlacesServlet extends HttpServlet {
 
+  /** {@link Type} of an {@link ArrayList} containing {@link String}, for gson decoding. */
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
+
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
```
zstewar1, line 43: I don't see anywhere where this logger is used here right now, remove?

byway/src/main/java/com/google/sps/data/UserInfo.java
Leo2452: I changed both entities to align with the class name, but since it's not a hard req I changed both class entities back.

Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Updated

Leo2452: Updated again: used logger when the trip is null on top of setting the response status code

PR #19, opened 2020-08-04
Class setups

Leo2452:
Set up Trip and User classes to format information for the user.


byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A class to make a Trip type, containing
+ * a specific id, destinations, interests and routes 
+ * to be used when updating a trip.
+ */
+public final class Trip{
+
+  private long id;
+  private ArrayList<String> destinations;
```
renau06, line 27: I think we need a start variable since we have that incldued as a Trip property in datstore schema

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public ArrayList<String> getRoute() {
+      return this.route;
+  }
+
+  public ArrayList<String> getDestinations() {
+    return this.destinations;
+  }
+}
```
renau06, line 64: Would we need get TripId?

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * A class to make a User type, containing an optional 
+ * email String and a list of trip IDs the user can have. 
+ * Users that are not logged in can only have one trip ID.
+ */
+public final class User{
+
+  private String email;
+  private ArrayList<long> tripIds;
```
renau06, line 27: throughout this file, the `ArrayList<long> `should be changed to `ArrayList<Long>`

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayList;
+
+/**
+ * A class to make a Trip type, containing
+ * a specific id, destinations, interests and routes 
+ * to be used when updating a trip.
+ */
+public final class Trip{
+
+  private long id;
```
renau06, line 26: I think we should save the id as a string because we will be able to call KeyFactory.stringToKey

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,53 @@
...
+    this.email = email;
+    this.tripIds = tripIds;
+  }
+
+  public User(String email) {
+    this.email = email;
+    this.tripIds = new ArrayList<>();
+  }
+
+  public ArrayList<String> getEmail() {
```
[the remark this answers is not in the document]
Leo2452: oops, fixed

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,64 @@
...
+              ArrayList<String> interests) {
+    configureTrip(id, destinations, interests, null);
+  }
+
+  public Trip(long id, ArrayList<String> destinations,
+              ArrayList<String> interests, ArrayList<String> route) {
+    configureTrip(id, destinations, interests, route);
+  }
+
+  private void configureTrip(long id, ArrayList<String> destinations,
```
j-vent: I think you can just rewrite this as a constructor, because if the user gives a null value for the last two arraylists(interests and route) the constructor will just assign those variables as null. I don't think it would cause any issues since our only primitive type (long) is never null.

byway/src/main/java/com/google/sps/data/User.java
Leo2452: Changed to String type, to store keys instead

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Good point, fixed

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,53 @@
...
+  private ArrayList<String> tripIds;
+
+  public User(String tripId) {
+    ArrayList<String> singleTrip = new ArrayList<>();
+    singleTrip.add(tripId);
+    this.email = "";
+    this.tripIds = singleTrip;
+  }
+
+  public User(String email, ArrayList<String> tripIds) {
```
j-vent, line 36: Not sure when this case would ever occur so maybe deleting it would be okay.

Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,60 @@
...
+    this.id = id;
+    this.start = start;
+    this.destinations = destinations;
+    this.interests = interests;
+    this.route = route;
+  }
+
+  public ArrayList<String> getInterests() {
+      return this.interests;
+  }
```
zstewar1, line 43: Be aware that if you return an ArrayList like this, even though you only have `get` methods, it will still be possible to modify the contents of this class, since the returned list is mutable and.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * A class to make a Trip type, containing
+ * a specific id, destinations, interests and routes 
+ * to be used when updating a trip.
+ */
+public final class Trip {
+
+  private String id;
+  private String start;
```
zstewar1, line 27: If you intend to allow these to be replaced, maybe add setters for them, otherwise mark them as `final`.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,60 @@
...
+  private ArrayList<String> destinations;
+  private ArrayList<String> interests;
+  private ArrayList<String> route;
```
zstewar1, line 30: Make these lists `final`. Note that if you intend to allow modification, it's still possible to change the contents of the list without replacing it.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public Trip(String id, String start, ArrayList<String> destinations,
+              ArrayList<String> interests, ArrayList<String> route) {
+    this.id = id;
+    this.start = start;
+    this.destinations = destinations;
+    this.interests = interests;
+    this.route = route;
```
zstewar1, line 38: Note that when writing container types it's often a bad idea to blindly trust that a collection provided by calling code won't be used elsewhere. Sharing the same list in different places by accident can lead to unintended issues, such as accidentally adding things to the same list multiple times. You may want to create a new list with a copy of the contents.

Note that if you use `this.interests = new ArrayList<>(interests)`, you can be more general in what type the constructor accepts, since you can construct an arraylist by [copying any `Collection`](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList-java.util.Collection-), you can make the arguments `Collection<T>` instead of `ArrayList<T>`, which gives the caller more choice of what type to pass in.

In general, it's a good idea to make the types of arguments you accept as general as reasonable, to provide the calling code with maximum flexibility around how they use your code.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,60 @@
...
+  private String id;
+  private String start;
+  private ArrayList<String> destinations;
+  private ArrayList<String> interests;
+  private ArrayList<String> route;
+
+  public Trip(String id, String start, ArrayList<String> destinations,
+              ArrayList<String> interests, ArrayList<String> route) {
+    this.id = id;
+    this.start = start;
```
zstewar1, line 35: Static import `com.google.common.base.Preconditions.checkNotNull` and wrap these assignments as appropriate:

`this.start = checkNotNull(start, "start");` (note the `"start"` is the error message that will appear if `start` is null. We usually just write the name of the parameter that was null when doing these kind of checks, since that's sufficient to figure out which thing caused the problem.)

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A class to make a User type, containing an optional 
+ * email String and a list of trip IDs the user can have. 
+ * Users that are not logged in can only have one trip ID.
+ */
+public final class User {
+
+  private String email;
+  private ArrayList<String> tripIds;
```
zstewar1, line 27: Notes about final, `checkNotNull` apply here as well.

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public User(String email, ArrayList<String> tripIds) {
```
zstewar1, line 29: You may want a userid string as well, since email addresses can change.

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: If it were modified elsewhere, after making a Trip class, wouldn't it leave the Trip containers unmodified since it passes in a copy? This makes it sound as if these collections would be passed by reference, which Java doesn't do. Maybe I'm wrong, since said parameters being passed would require a new keyword to make an object, which would likely be using the same object (in which case what you said would apply). Not sure how Java works internally so would appreciate some insight/clarification.

Also, fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public final class Trip {
+
+  private final String id;
+  private final String start;
+  private final ArrayList<String> destinations;
+  private final ArrayList<String> interests;
+  private final ArrayList<String> route;
+
+  public Trip(String id, String start, Collection<String> destinations,
```
landondare, line 34: It would also be nice to have a constructor that takes in an entity object returned from datastore and constructs a Trip. Or you could have a separate static helper function that builds a Trip from a datastore return object. Something like 
```java
createTripFromEntity(Entity tripEntity){
  id = TripEntity.getProperty("id");
  ...
  return new Trip(id, start, destinations, ....);
}
```

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,68 @@
...
+    Collection<String> validDestinations = checkNotNull(destinations, "destinations");
+    Collection<String> validInterests = checkNotNull(interests, "interests");
+    Collection<String> validRoute = checkNotNull(route, "route");
+    this.destinations = new ArrayList<String>(validDestinations);
+    this.interests = new ArrayList<String>(validInterests);
+    this.route = new ArrayList<String>(validRoute);
+  }
+
+  public ArrayList<String> getInterests() {
+    ArrayList<String> interestsCopy = this.interests;
```
landondare, line 47: This is not actually returning a different list. It returning a reference to the same list. That may or not be okay depending on how you plan on using this. If you actually want to make a copy to return do this:
```
ArrayList<String> interestsCopy = new ArrayList<>(this.interests);
```

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,68 @@
...
+    return interestsCopy;
+  }
+
+  public ArrayList<String> getRoute() {
+    ArrayList<String> routeCopy = this.route;
```
landondare, line 52: Same

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,68 @@
...
+    return routeCopy;
+  }
+
+  public ArrayList<String> getDestinations() {
+    ArrayList<String> destinationsCopy = this.destinations;
```
landondare, line 57: Same

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,51 @@
...
+    Collection<String> validTripIds = checkNotNull(tripIds, "tripIds");
+    this.tripIds = new ArrayList<>(validTripIds);
+  }
+
+  public String getEmail() {
+    return this.email;
+  }
+
+  public ArrayList<String> getTripIds() {
+    ArrayList<String> tripIdsCopy = this.tripIds;
```
landondare, line 44: Same

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.ArrayList;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A class to make a Trip type, containing
+ * a specific id, destinations, interests and routes 
+ * to be used when updating a trip.
+ */
+public final class Trip {
+
```
landondare, line 27: Consider storing the datastore entity KIND on this class as well as a static variable

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,51 @@
...
+ * email String and a list of trip IDs the user can have. 
+ * Users that are not logged in can only have one trip ID.
+ */
+public final class User {
+
+  private String email;
+  private String userId;
+  private ArrayList<String> tripIds;
+
+  public User(String email, String userId, Collection<String> tripIds) {
```
landondare, line 32: Again, would be nice to have an additional constructor that takes in a User object from datastore and extracts this information.

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Noted. I created the helper functions within each class along with checks to make sure that
a) the entity passed is is not null itself, and
b) each property required to make a User/Trip class was valid/not null

Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed (for all cases I wanted to return a new list)

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+  private final String id;
+  private final String start;
+  private final ArrayList<String> destinations;
+  private final ArrayList<String> interests;
+  private final ArrayList<String> route;
+  private final static String KIND = "Trip";
```
landondare, line 35: Make this public so other classes and can reference it and give it a more descriptive name. Something that specifies that it deals with datastore. Maybe: `DATASTORE_ENTITY_KIND` or `DATASTORE_ENTITY`

byway/src/main/java/com/google/sps/data/Trip.java
landondare: Nice

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+    this.start = checkNotNull(start, "start");
+    Collection<String> validDestinations = checkNotNull(destinations, "destinations");
+    Collection<String> validInterests = checkNotNull(interests, "interests");
+    Collection<String> validRoute = checkNotNull(route, "route");
+    this.destinations = new ArrayList<String>(validDestinations);
+    this.interests = new ArrayList<String>(validInterests);
+    this.route = new ArrayList<String>(validRoute);
+  }
+
+  public Trip createTripFromEntity(Entity tripEntity) {
```
landondare, line 49: Make this `public static` and move to the bottom of the file. It needs to be static so you can reference it without having to first create a Trip object. Try to group static methods away from non-static methods (hence move it to the bottom).

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,67 @@
...
+  private final static String KIND = "user";
+
+  public User(String email, String userId, Collection<String> tripIds) {
+    this.email = checkNotNull(email, "email");
+    this.userId = checkNotNull(userId, "userId");
+    Collection<String> validTripIds = checkNotNull(tripIds, "tripIds");
+    this.tripIds = new ArrayList<>(validTripIds);
+  }
+
+  public User createUserFromEntity(Entity userEntity) {
```
landondare, line 42: Same as above, make this `public static` and move to the bottom of the file.

byway/src/main/java/com/google/sps/data/Trip.java
landondare: Nice!

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private String email;
+  private String userId;
+  private ArrayList<String> tripIds;
+  public static final String DATASTORE_ENTITY_KIND = "user";
+
+  public User(String email, String userId, Collection<String> tripIds) {
+    this.email = checkNotNull(email, "email");
+    this.userId = checkNotNull(userId, "userId");
+    Collection<String> validTripIds = checkNotNull(tripIds, "tripIds");
```
zstewar1, line 38: Same as above with uses of checkNotNull.

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,67 @@
...
+    this.tripIds = new ArrayList<>(validTripIds);
+  }
+
+  public String getEmail() {
+    return this.email;
+  }
+
+  public ArrayList<String> getTripIds() {
+    return new ArrayList<String>(this.tripIds);
```
zstewar1, line 47: Same as above, either use `Collectoins.unmodifiableList` or allow mutation.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+ * to be used when updating a trip.
+ */
+public final class Trip {
+
+  private final String id;
+  private final String start;
+  private final ArrayList<String> destinations;
+  private final ArrayList<String> interests;
+  private final ArrayList<String> route;
+  public static final String DATASTORE_ENTITY_KIND = "Trip";
```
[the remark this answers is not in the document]
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+  public String getTripId() {
+    return this.id;
+  }
+
+  public String getStart() {
+    return this.start;
+  }
+
+  public static Trip createTripFromEntity(Entity tripEntity) {
+    Entity validEntity = checkNotNull(tripEntity, "Trip entity is null");
```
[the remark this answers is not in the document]
Leo2452: I added it here because the entity itself could be null. In the other cases, if it tries to get a property and that property itself is null, it would return null because it's missing the property. But, if the entity itself is null and it tries to get a property, it'd return null for the wrong reason since it'd indicate a property was missing. I removed it just in case I'm overthinking it, but that was my rationale behind that check

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+    Collection<String> validDestinations = checkNotNull(destinations, "destinations");
+    Collection<String> validInterests = checkNotNull(interests, "interests");
+    Collection<String> validRoute = checkNotNull(route, "route");
+    this.destinations = new ArrayList<String>(validDestinations);
+    this.interests = new ArrayList<String>(validInterests);
+    this.route = new ArrayList<String>(validRoute);
+  }
+
+  public ArrayList<String> getInterests() {
+    return new ArrayList<String>(this.interests);
```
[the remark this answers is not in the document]
Leo2452: Not sure if I understand. How could the inner list change and this unmodifiable list would show the updated values, since it's unmodifiable and changing the inner list means modifying it? Maybe I'm misinterpreting which list is the inner list, Is the inner list the list within the class? If so, we wouldn't ever be able to modify (as of now with no setters) unless we create a new Trip, in which case it is trivial. If the inner list is the list created before the Trip class is made, I'm assuming the whole point of having the unmodifiable list this way means the changed won't be made inside the class, even if changed after it were declared.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public String getTripId() {
+    return this.id;
+  }
+
+  public String getStart() {
+    return this.start;
+  }
+
+  public static Trip createTripFromEntity(Entity tripEntity) {
```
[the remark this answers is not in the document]
Leo2452: I see, fixed the repetition.

byway/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * A class to make a User type, containing an optional 
+ * email String and a list of trip IDs the user can have. 
+ * Users that are not logged in can only have one trip ID.
+ */
+public final class User {
+
+  private String email;
+  private String userId;
+  private ArrayList<String> tripIds;
```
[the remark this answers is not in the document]
Leo2452: Forgot to fix previously, updated now

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,88 @@
...
+  private final ArrayList<String> route;
+  public static final String DATASTORE_ENTITY_KIND = "Trip";
+
+  public Trip(String id, String start, Collection<String> destinations,
+              Collection<String> interests, Collection<String> route) {
+    this.id = checkNotNull(id, "id");
+    this.start = checkNotNull(start, "start");
+    Collection<String> validDestinations = checkNotNull(destinations, "destinations");
+    Collection<String> validInterests = checkNotNull(interests, "interests");
+    Collection<String> validRoute = checkNotNull(route, "route");
```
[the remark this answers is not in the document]
Leo2452: I added checkNotNull  to see if the params passed in themselves were null. I thought it was overkill, but wasn't sure when/where to check for null values, so did it for the collections as well. If I removed this, would I just do
`this.interests = new ArrayList<>(interests)`? It wouldn't check if the collection is null, which would make a new arraylist but with null value(s).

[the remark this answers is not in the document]
zstewar1: You can just call `checkNotNull` without doing anything with the returned value, the returned value doesn't matter. The function works roughly like this:

```java
static <T> T checkNotNull(T value, String failureMessage) {
  if (value == null) {
    throw new NullPointerException(failureMessage);
  }
  return value;
}
```

so if the value is null, you are guaranteed an exception, and the next line will not execute. If the value is not null, it is returned unmodified. Returning the same value unmodified is just a convenience feature which is there to allow you to write a null check in a single line when that's appropriate, like with the `start` and `id` where you write `this.id = checkNotNull(id, "id");`. when it's not appropriate to write the check in a single line, you can write `checkNotNull(destinations, "destinations");` on one line without using the returned value, and just assume that the value will not be null from then on (because if it was, you would have thrown an exception.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,89 @@
...
+    this.destinations = new ArrayList<String>(validDestinations);
+    this.interests = new ArrayList<String>(validInterests);
+    this.route = new ArrayList<String>(validRoute);
+  }
+
+  public ArrayList<String> getInterests() {
+    return (ArrayList<String>) Collections.unmodifiableList(this.interests);
```
zstewar1, line 52: The value of `unmodifiableList` is not an `ArrayList`, they are different types.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public String getTripId() {
+    return this.id;
+  }
+
+  public String getStart() {
+    return this.start;
+  }
+
+  public static Trip FromEntity(Entity tripEntity) {
```
zstewar1, line 71: Keep the casing consistent. Methods should use `lowerCamelCase`.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,89 @@
...
+    checkArgument(tripEntity.getKind().equals(DATASTORE_ENTITY_KIND),
+      "Wrong Entity kind. Expected %s, received %s", DATASTORE_ENTITY_KIND, tripEntity.getKind());
+    String id = checkNotNull((String) tripEntity.getProperty("id"),
```
zstewar1, line 74: If you want something to be the canonical "ID" of an entity, it would be good to make it the Key of the entity, that way you can be sure you can't accidentally have to multiple entities with the same ID.

Note that for serialization to JSON, you can still store the "id" as a string, but you might find it useful to have a
```java
Key getKey() {
   return new Key(DATASTORE_ENTITY_KIND, id);
}
```

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Updated

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Oops, just deleted `createTrip` without fixing this, fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: We decided to rename this id variable throughout the app to reference the trip key itself rather than labelling it as id, since we could mistakenly be referring to the key's id. I implemented this function as well but with a `stringToKey` method instead, effectively doing the same thing.

Updated

zstewar1: There is a specific field on the Entity for its key:

https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Entity.html#getKey--

If you use the actual datastore key of the entity, you are guaranteed that no other entity has the same key as this entity. If you store a string in your own property, you could have multiple entities with the same value for `tripEntity.getProperty("id")`.

Using `id = KeyFactory.keyToString(tripEntity.getKey())` is fine with me if you want to have the ID just be the encoded key value and that's probably easier.

However, what I was suggesting above was that you can also use something like `id = tripEntity.getKey().getName()`, assuming when you create the entity you get the key using `createKey(DATASTORE_ENTITY_KIND, theIdYouWantForThisEntity)`. Then whenever you get a request with a tripId, you turn the tripId back into a datastore key the same way, using `createKey(DATASTORE_ENTITY_KIND, tripId)`.

This approach, using `createKey(DATASTORE_ENTITY_KIND, id)` is especially useful for the `User` type, since you would be getting a User from AppEngine, and can just use `createKey(DATASTORE_ENTITY_KIND, user.getUserId())` as the key for the entity, re-using the [ID provided by AppEngine](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html#getuserid) as the ID in datastore.

byway/src/main/java/com/google/sps/data/Trip.java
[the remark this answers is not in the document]
zstewar1: Yes, having this check is fine, but there is no need to assign the result of `checkNotNull` to a new variable. Here and in `User`.

byway/src/main/java/com/google/sps/data/Trip.java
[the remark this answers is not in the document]
zstewar1: `unmodifiableList` returns a new `List` object with a reference to the original list inside it. When you do any non-modifying operation like `get` it just calls `get` on the inner list. If you try to call any method that would modify the original list, like `add`, it throws an exception. It's still possible to call `add` on the arraylist itself, but the wrapper list prevents modification through that wrapper.

The code for this is public and not actually that hard to read: https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collections.java#l1355

Notice how it keeps a reference to the original list: `final List<? extends E> list;`, and look at the implementations for `get` and `set`:

```java
        public E get(int index) {return list.get(index);} // delegates this operation to the inner list.
        public E set(int index, E element) { // rejects if you try to set anything.
            throw new UnsupportedOperationException();
        }
```

Since this is just a wrapper object around another list, the inner list is still perfectly usable and still supports operations that add things to it, as long as you call `add` on a reference to the original list that you've saved elsewhere, like `this.interests`, but will not allow additions or changes through the wrapper class.

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Both ways are implementable/relatively straight forward. I first set it up as passing the key's id, converting to long, and creating a key, `createKey(DATASTORE_ENTITY_KIND`, id)`, and it worked fine. But it felt like an extra step, converting strings to longs and checking for number format exceptions, then creating the key. With the other method, it only converts two-way, string to key, and the parameters are passed in as strings so it avoids getting the correct type.

This approach applies to both the Trip and User class. I think `createKey()` required the id of the key itself rather than `user.getUserId()`. I'm not sure of the distinction between both, since the key's id references the user, but so should userId which is based on the same user. Also, wouldn't the appengine `User` class interfere with the custom `User` class being made here? For now, I'll implement it with the keyToString method rather than use id, and can update it with ids if problems arise

byway/src/main/java/com/google/sps/data/Trip.java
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/java/com/google/sps/data/Trip.java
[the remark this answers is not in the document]
Leo2452: I see, makes more sense - thanks

byway/src/main/java/com/google/sps/data/Trip.java
zstewar1: When you create a new entity, it will always have *a* key, and you have 3 choices for how to decide on a key for it.

1) Create an entity with a Kind only. When you insert it into the datastore, it will get a numeric (`long`) ID value chosen automatically. To do this, you construct the entity with `new Entity(DATASTORE_ENTITY_KIND)` and then when you call `datastore.put(entity)` the key will be assigned, and can be retrieved with `entity.getKey()`.
2) Choose your own value for the `long` ID. You create your Key ahead of time via the `KeyFactory`, and create the entity with it. Example: `new Entity(KeyFactory.createKey(DATASTORE_ENTITY_KIND, yourChosenId))`. When you use `datastore.put(entity)`, the ID will be left with the value you chose.
3) Choose your own value for the `String` `name` of the key, and create the entity with it. Example: `new Entity(KeyFactory.createKey(DATASTORE_ENTITY_KIND, yourChosenName))`. When you use `datastore.put(entity)`, the `name` will be used, and no numeric `id` will be assigned.

Note that when you are using option (2) or (3), you have to make sure that the number or string that you choose really is unique, otherwise you will replace an existing entity with that key when you use `datastore.put(entity)`, rather than add a new entity.

Option (3) is useful for working with the AppEngine `UserService`, because the `UserService` provides a unique ID per-user already, so it's pretty straightforward to just use that as your the Key's `name` rather than having the datastore choose a numeric `id` for you. You'll find that this is especially convenient when it comes to retrieving the user entity for a particular user. Example below, but first, let me address this:

> Also, wouldn't the appengine User class interfere with the custom User class being made here?

No, but also yes, sort of.

In Java, your classes have both a name and a package. The full name of a Java class is really the combination of the package name and the class name. Since your `User` class is declared in a file with `package com.google.sps.data`, the full name of the `User` class is actually `com.google.sps.data.User`. Meanwhile, the `User` class from AppEngine is actually `com.google.appengine.api.users.User`. When you write `import com.google.appengine.api.users.User`, what you're saying is "when I write `User`, what I mean is `com.google.appengine.api.users.User`". Imports are technically optional! You could *technically* write all of your code using only the full names of classes including the package. Imports exist primarily to make code easier to read and write for humans. Also, if you are writing a class in a particular package, any other classes in that package are available by default as if you imported them.

So what this means is that *technically*, the `com.google.appengine.api.users.User` will not interfere with your `User` class, because you can write `User` when you mean `com.google.sps.data.User`, and then spell-out the full `com.google.appengine.api.users.User`. But practically speaking, yeah this kind of interferes in your code, since now you have to write a really long name when using one of them. So practically speaking, it might be a good idea to rename your `User` class to something else, like `UserInfo`, so you can have both names imported at the same time.

----

Anyway, back to how you would use the user ID as a key.

You can imagine writing code that does something sort of like this in order to get or create the User entity:

```java
com.google.appengine.api.users.User user = userService.getCurrentUser();
if (user != null) {
  Key userKey = KeyFactory.createKey(DATASTORE_ENTITY_KIND, user.getUserId());
  Entity userEntity;
  try {
    userEntity = datastore.get(userKey);
  } catch (EntityNotFoundException ex) {
    userEntity = new Entity(userKey);
    userEntity.setProperty("email", user.getEmail());
    userEntity.setProperty("tripIds", new ArrayList<Key>()); // note ArrayList<Key>; when referring to datastore entites from other datastore entities, I'd recommend using explicit Key rather than a string formatted value.
    datastore.put(userEntity);
  }
  return fromEntity(userEntity);
}
return null;
```

Leo2452: I see, so the userId would make sense since we can input it directly for the (now changed) `UserInfo` class. I was confused about using both User classes if we imported them to the same place since we could get errors or stuff would be overwritten. They do have different constructors, so maybe it would have been accepted? Not sure, either way changed for the sake of readability and to save potential confusion. The set up makes more sense now, thanks

Fixed

byway/src/main/java/com/google/sps/data/UserInfo.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+
+/** A class to make a UserInfo type, containing an email, id and a list of trip IDs per user. */
+public final class UserInfo {
+
+  public static final String DATASTORE_ENTITY_KIND = "user";
+
+  private final String email;
+  private final String id;
+  private final ArrayList<Key> tripIds;
```
zstewar1, line 35: Hm, storing Keys in datastore is probably better for querying the datastore, but worse for converting them with `gson`, since you don't really know how they'll end up being converted to gson.

If you want to use the same objects for gson serialization as for working with datastore entities, then it's probably OK to store them as strings.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public static Trip fromEntity(Entity tripEntity) {
+    checkNotNull(tripEntity, "tripEntity");
+    checkArgument(
+        tripEntity.getKind().equals(DATASTORE_ENTITY_KIND),
+        "Wrong Entity kind. Expected %s, received %s",
+        DATASTORE_ENTITY_KIND,
+        tripEntity.getKind());
+    String id =
+        checkNotNull((String) tripEntity.getProperty("id"), "Trip entity does not contain an id");
```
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/java/com/google/sps/data/UserInfo.java
```diff
@@ -0,0 +1,79 @@
...
+    checkNotNull(userEntity, "User entity is null");
+    checkArgument(
+        userEntity.getKind().equals(DATASTORE_ENTITY_KIND),
+        "Wrong entity kind. Expected %s, received %s",
+        DATASTORE_ENTITY_KIND,
+        userEntity.getKind());
+    String email =
+        checkNotNull(
+            (String) userEntity.getProperty("email"), "User entity does not contain an email");
+    String id =
```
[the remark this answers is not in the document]
Leo2452: Renamed to userId to save the String rather than the Key to make it simpler to send to front end when using gson/json, as mentioned below.

byway/src/main/java/com/google/sps/data/UserInfo.java
Leo2452: Updated as Strings. We won't have to query the datastore, since we will retrieve the stringified key from the front end and can access the same entity directly.

byway/src/main/java/com/google/sps/data/Trip.java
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public static Trip fromEntity(Entity tripEntity) {
+    checkNotNull(tripEntity, "tripEntity");
+    checkArgument(
+        tripEntity.getKind().equals(DATASTORE_ENTITY_KIND),
+        "Wrong Entity kind. Expected %s, received %s",
+        DATASTORE_ENTITY_KIND,
+        tripEntity.getKind());
+    String keyString =
+        checkNotNull(
```
zstewar1, line 90: I don't think the `Key` can be null, so you don't need to check it.

byway/src/main/java/com/google/sps/data/UserInfo.java
```diff
@@ -0,0 +1,79 @@
...
+    checkNotNull(userEntity, "User entity is null");
+    checkArgument(
+        userEntity.getKind().equals(DATASTORE_ENTITY_KIND),
+        "Wrong entity kind. Expected %s, received %s",
+        DATASTORE_ENTITY_KIND,
+        userEntity.getKind());
+    String email =
+        checkNotNull(
+            (String) userEntity.getProperty("email"), "User entity does not contain an email");
+    String userId =
```
zstewar1, line 70: For users, you should use the `userId` provided by UserService as the Key's name. That means when retrieving the `userId` from the entity, you should use `userEntity.getKey().getName()`.

byway/src/main/java/com/google/sps/data/UserInfo.java
```diff
@@ -0,0 +1,79 @@
...
+  public List<String> getTripIds() {
+    return Collections.unmodifiableList(this.tripIds);
+  }
+
+  public String getUserId() {
+    return this.userId;
+  }
+
+  public Key getKey() {
+    return KeyFactory.stringToKey(this.userId);
```
zstewar1, line 57: For users, you should use the `userId` provided by UserService as the Key's name. That means when reconstructing the `Key` from the `userId`, you should use `KeyFactory.create(DATASTORE_ENTITY_KIND, userId)`.

`stringToKey` is only for strings created by `keyToString`.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,110 @@
...
+    this.start = checkNotNull(start, "start");
+    checkNotNull(destinations, "destinations");
+    checkNotNull(interests, "interests");
+    checkNotNull(route, "route");
+    this.destinations = new ArrayList<String>(destinations);
+    this.interests = new ArrayList<String>(interests);
+    this.route = new ArrayList<String>(route);
+  }
+
+  public List<String> getInterests() {
```
zstewar1, line 58: Can you add comments to these methods to describe what they are? Not just like "the user's interests", but rather it would be good to include information about the provenance of the data and what format it is in. E.g. "Interests the user selected for the trip, as free-form, user-supplied text."

Applies to all getters for both data classes here.

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Updated

byway/src/main/java/com/google/sps/data/UserInfo.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/UserInfo.java
Leo2452: Fixed

byway/src/main/java/com/google/sps/data/Trip.java
Leo2452: Updated

PR #20, opened 2020-08-05
Route page selection backend

j-vent:
These are the servlets I used to get the third page logic working. I think RecsServlet may be removed in the final implementation since its functionality is just setting up a Trip (done in first page) and also sending recommendations to the js (done in the second page).
Trip.java was also a simplification of what we will have in the final version. 
StopServlet.java is needed to update the destinations property of the Trip.


byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,23 +42,41 @@
...
-    Gson gson = new Gson();
-    String json = gson.toJson(stops);
-    
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Replace ID filtering with .get(Key)
+    // Duplicate code was just kept for this reason
+    Filter propertyFilter = new FilterPredicate("id", FilterOperator.EQUAL, 1234);
+    Query query = new Query(Trip.KIND).setFilter(propertyFilter);
+    PreparedQuery results = datastore.prepare(query);
+    List<String> stops= new ArrayList<>();
```
renau06, line 57: Change List to ArrayList

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,23 +42,41 @@
...
+
+    // Should only be one entity with this ID
+    for(Entity entity: results.asIterable()){
```
renau06, line 60: Do you need a forloop if only one entity will have that ID? I'm not exactly sure how it works so its possible that you might still need the forloop

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,23 +42,41 @@
...
-    String stop = request.getParameter("text");
-    stops.add(stop);
-  }
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException{
+    String stopsAsJSON = request.getParameter("stops");
+
+    ArrayList<String> stops = gson.fromJson(stopsAsJSON, ARRAYLIST_STRING);
+    Filter propertyFilter = new FilterPredicate("id", FilterOperator.EQUAL, 1234);
+    Query query = new Query(Trip.KIND).setFilter(propertyFilter);
+    PreparedQuery results = datastore.prepare(query);
```
Leo2452, line 75: Note that if you know the entity's kind and the key's id, you can create a key, `KeyFactory.createKey(KIND, id)`, and use `datastore.get(key)` to retrieve the exact entity without having to use queries and filters.

Same goes for the doGet function using the queries as well

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,23 +42,41 @@
...
+
+    for(Entity entity: results.asIterable()){
+      entity.setProperty("destinations", stops);
+      datastore.put(entity);
+    }
```
Leo2452, line 80: The comment above is also for the purpose of not iterating through all results, since it should only find one unique trip. I have an implementation using this in my `store-interests` branch using the hard-coded id value for the trip if you want to take a look: https://github.com/googleinterns/step200-2020/pull/18/commits/9a96386d9b991e42e3eb325d83c44317756ab56c

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: I think both implementations are fine, because ArrayLists are just a specific implementation of the List interface.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Oh gotcha, thanks. I'll change both functions to use that instead.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Yeah I'm just not familiar with handling the type `PreparedQuery` of results so I chose an inefficient solution. Will look into possible fixes!

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Yeah, my only issue with that is the hardcoded ID we made is not used, we have to still rely on the randomly generated ID generated at creation of the entity. But I don't think this is a huge deal since we are planning to use the Trip keys directly, so I will change for now.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -0,0 +1,40 @@
...
+ * TODO: Replace with actual Trip.java class
+ */ 
+public final class Trip{
+  private final long id;
+  private final ArrayList<String> destinations;
+  public static final String KIND = "Trip";
+  
+  public Trip(long id, ArrayList<String> destinations) {
+    this.id = id;
+    this.destinations = destinations;
```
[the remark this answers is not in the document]
j-vent: Okay, thank you for the detailed explanation, that makes sense. Changed!

byway/src/main/java/com/google/sps/servlets/RecsServlet.java
```diff
@@ -38,34 +39,22 @@
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Query query = new Query(Recommendation.KIND); 
-    PreparedQuery results = datastore.prepare(query);
-    List<Recommendation> recs = new ArrayList<>();
-
-    for (Entity entity: results.asIterable()){
-      recs.add(Recommendation.fromEntity(entity));
-    }
+    List<String> recs = new ArrayList<String>(Arrays.asList("Times Square", "MOMA", "Central Park" ));
```
zstewar1, line 56: If you don't need to be able to modify the list, you can just directly use `Arrays.asList` as a `List`, rather than creating an `asList` and then making a copy of it to create an `ArrayList`.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -14,35 +14,66 @@
...
-    Gson gson = new Gson();
-    String json = gson.toJson(stops);
-    
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> stops= new ArrayList<>();
+    try{
+      Entity entity = datastore.get(KeyFactory.createKey(Trip.KIND, 5910974510923776L));
+      stops = (ArrayList<String>) entity.getProperty("destinations");
+    } catch (EntityNotFoundException e){
+      System.out.println("Could not retrieve stops due to faulty key " + e);
```
zstewar1, line 59: https://google.github.io/flogger/

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -14,35 +14,66 @@
...
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
 
+  /* Passes saved destinations stops (if any) to be shown in the schedule panel */
   @Override
-  public synchronized void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Gson gson = new Gson();
-    String json = gson.toJson(stops);
-    
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> stops= new ArrayList<>();
```
zstewar1, line 54: If you want a default empty list which you won't need to be able to append to, you can use `Collections.emptyList()`, rather than constructing a new empty `ArrayList`. This avoids allocation, because there is only one instance of `Collections.emptyList()` for the entire program. (If you do need to `add` or `remove` anything, that won't work, but if you just want to put an empty list in a variable, and aren't going to `add` or `remove` (replacing the thing with a different list is fine of course), you can just use `emptyList()`

byway/src/main/java/com/google/sps/servlets/RecsServlet.java
j-vent: Yup, these are just hardcoded values. Changed.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: I think I'll use emptyList since I am replacing the list with the destinations list in datastore. Thanks for explaining.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Added.

byway/src/main/java/com/google/sps/servlets/RecsServlet.java
```diff
@@ -35,34 +31,21 @@
...
-    Entity recEntity3 = new Entity(Recommendation.KIND);
-    recEntity1.setProperty("placename", "Times Square");
-    datastore.put(recEntity1);
-    recEntity2.setProperty("placename", "MOMA");
-    datastore.put(recEntity2);
-    recEntity3.setProperty("placename", "Central Park");
-    datastore.put(recEntity3);
-  }
+  /* Intializes a mock Trip entity
+  public void init(){
```
landondare, line 35: Seems you can remove this comment.

j-vent: Removed.

byway/pom.xml
```diff
@@ -49,19 +49,6 @@
...
-      <artifactId>appengine-api-1.0-sdk</artifactId>
-      <version>1.9.59</version>
-    </dependency>
-    
-    <!-- https://mvnrepository.com/artifact/com.google.flogger/flogger -->
-    <dependency>
-      <groupId>com.google.flogger</groupId>
-      <artifactId>flogger</artifactId>
-      <version>0.5.1</version>
-    </dependency>
```
zstewar1, line 63: Why did you remove this?

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +36,35 @@
...
 @WebServlet("/api/stop")
 public final class StopsServlet extends HttpServlet {
-  @GuardedBy("this")
-  private ArrayList<String> stops = new ArrayList<String>();
-
   private final Gson gson = new Gson();
-  // Implement datastore in routepage branch
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
```
zstewar1, line 42: Try to keep static fields separate from non-static fields. Usually it's good to put statics first in the class (since they're shared across all instances, it's nice to show them first) and put a blank line after them to separate them from non-static field.s

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +36,35 @@
...
   @Override
-  public synchronized void doGet(HttpServletRequest request, HttpServletResponse response)
-      throws IOException {
-    Gson gson = new Gson();
-    String json = gson.toJson(stops);
-
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> stops = Collections.emptyList();
+    try {
+      Entity entity = datastore.get(KeyFactory.createKey(Trip.KIND, 5910974510923776L));
```
zstewar1, line 49: Hardcoded value is likely only valid for your local datastore. If doing temp hardcoded stup, it woudl be good to explicitly set the key of the entity in the `init` that creates the entity so that it's alway the same across all datastore instances.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +36,35 @@
...
+      stops = (ArrayList<String>) entity.getProperty("destinations");
+    } catch (EntityNotFoundException e) {
+      logger.atInfo().withCause(e).log("Could not retrieve stops due to faulty key", e);
```
zstewar1, line 52: Note that you have no `%s` formatting parameters in the logged message, so the `e` being passed at the end won't actually be formatted into the message. Also note that you're already using `withCause(e)` which will appropriately include the cause in the error, so you don't really need to format the error into the message string.

For the message string, try to be more specific, e.g. if reading this message without being familiar with this code what does "faulty key" mean? Consider something like `.log("Could not retrieve Entity for Trip with key %s while trying to get the stops", key)` (note that to do this you need to store the result of `KeyFactory.createKey` before the `try` block). That way if this error happens, you can definitely see what key was *actually* provided and have an easier time debugging. This applies to other places you have log messages as well. Be as specific as you reasonably can, so it's easy to tell:

- What happened ("Could not retrieve Entity for Trip")
- What data it affected ("with key %s")
- What you were trying to do ("while trying to get the stops")

This kind of context can make it a lot easier to tell why something caused an issue, or even make it possible to tell right away when what you actually did doesn't match what you were trying to do; for example, if `doPost` says "while trying to *update* the stops", you can tell immediately whether the error was in the post request or the get request.

byway/src/main/java/com/google/sps/servlets/RecsServlet.java
j-vent: Huh, I changed this in commit df9a43e3dd6b7596c5f2fe925f89951d9bb8b70f.

byway/pom.xml
j-vent: It was one of the duplicate dependencies.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Oh okay, thanks for pointing this out I just lumped all final fields together. Changed.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Yup I initially used the ID I created in init() (1234) in place of the Long. However createKey() uses the random-generated ID which I have not figured out how to override, so I just found took this ID value from the datastore admin page and used it.

zstewar1: When you create a new entity to add to datastore, you have 3 choices for how to create the `Key` for it:

1) Create an entity with a Kind only. When you insert it into the datastore, it will get a numeric (`long`) ID value chosen automatically. To do this, you construct the entity with `new Entity(KIND)` and then when you call `datastore.put(entity)` the key will be assigned, and can be retrieved with `entity.getKey()`.
2) Choose your own value for the `long` ID. You create your Key ahead of time via the `KeyFactory`, and create the entity with it. Example: `new Entity(KeyFactory.createKey(KIND, yourChosenId))`. When you use `datastore.put(entity)`, the ID will be left with the value you chose.
3) Choose your own value for the `String` `name` of the key, and create the entity with it. Example: `new Entity(KeyFactory.createKey(KIND, yourChosenName))`. When you use `datastore.put(entity)`, the `name` will be used, and no numeric `id` will be assigned.

So if you have your 'init' setup the entities with the same key value always, you'll be creating them under that key, and can safely re-use the key elsewhere.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: I didn't notice I took out the format param and key when I switched to logger. I will try to be more detailed with my error messages, so thanks for breaking down what exactly needs to go into the error message.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Oh I see, I think I setup ID like how they told us in the walkthroughs where you make your own custom id field. I think the second method is best. I'll add my init() function again and see.

byway/pom.xml
zstewar1: These are not duplicates, both are needed. Note the `artifactId` is different:

```xml
    <!-- https://mvnrepository.com/artifact/com.google.flogger/flogger -->
    <dependency>
      <groupId>com.google.flogger</groupId>
      <artifactId>flogger</artifactId>
      <version>0.5.1</version>
    </dependency>

    <!-- https://mvnrepository.com/artifact/com.google.flogger/flogger -->
    <dependency>
      <groupId>com.google.flogger</groupId>
      <artifactId>flogger-system-backend</artifactId>
      <version>0.5.1</version>
    </dependency>
```

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +37,39 @@
...
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    ArrayList<String> stops = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    try {
+      Entity entity = datastore.get(key);
+      entity.setProperty("destinations", stops);
+      datastore.put(entity);
+    } catch (EntityNotFoundException e) {
+      logger.atInfo().withCause(e).log(
+          "Could not retrieve Entity for Trip with key %s while trying to update the stops", key);
+    }
```
zstewar1, line 73: It would be good to set status codes on success and failure, so you give the client code some idea what the error is, or what the result is. With `response.setStatus`, you can use the status code constants from `HttpServletResponse`, for example when the entity is not found, you could send `SC_NOT_FOUND`, and when a request succeeds but you have nothing to actually send back, you can send `SC_NO_CONTENT`.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +37,39 @@
...
+  /* Modifies the destinations array of Trip entity in datastore */
   @Override
-  public synchronized void doPost(HttpServletRequest request, HttpServletResponse response)
-      throws IOException {
-    String stop = request.getParameter("text");
-    stops.add(stop);
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    ArrayList<String> stops = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    try {
+      Entity entity = datastore.get(key);
```
zstewar1, line 67: When you make a `try` block, try to make it as narrowly scoped as you can. That way you don't accidentally catch errors from parts of the code that you didn't mean to handle. This is important for debugging, because if you catch errors you didn't intend to you may not realize what the actual source of an error is, or you can introduce bugs, for example if you handle `EntityNotFound` by creating the requested trip entity, but it turns out you actually caught `EntityNotFound` from code that was trying to get the User, you might overwrite an existing trip by accident.

byway/pom.xml
j-vent: Oh thanks for pointing that out, I didn't look close enough. Added back.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Okay, I've added this setStatus() to both methods.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Forgot to comment, done.

byway/pom.xml
```diff
@@ -42,6 +42,12 @@
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.8.6</version>
+    </dependency>
```
zstewar1, line 50: Looks like gson is already included in the version from master. You should probably just reset this file back to the way it is in master (with `git checkout master byway/pom.xml` while on this branch).

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -26,26 +37,49 @@
 /** Servlet that handles data for stops */
 @WebServlet("/api/stop")
 public final class StopsServlet extends HttpServlet {
-  @GuardedBy("this")
-  private ArrayList<String> stops = new ArrayList<String>();
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+  private static final int SC_NO_CONTENT = 204;
+  private static final int SC_NOT_FOUND = 404;
```
zstewar1, line 43: There are already constants for these declared in `HttpServletResponse`, just use those.
https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletResponse.html#SC_NOT_FOUND

byway/src/main/webapp/routepage.js
```diff
@@ -1,175 +0,0 @@
...
-    // TODO: add to recs later for better visuals on html
-    stops.push(rec);
-    updateStops();
-  });
-  return recBtn;
- }
-
-/* exported initMap */
-/* global google */
-
```
zstewar1, line 175: Do you want this file to be deleted from master? If not you should commit back the version of the file in master, by fetching it with `git checkout master byway/src/main/webapp/routepage.js` and then committing that.

byway/pom.xml
j-vent: Okay, I ran that command. The online merge tool just caused more conflicts.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
j-vent: Right, I just saw the list of status codes and numbers and decided to go off that instead.  Changed.

byway/src/main/webapp/routepage.js
j-vent: Will add back, and delete the Recommendation.java file that I don't need anymore but is in master.

PR #23, opened 2020-08-10
Route Selection waypoints

j-vent:
When a stop is added or removed, the map is redrawn to account for these changes.


byway/src/main/webapp/routepage.js
```diff
@@ -70,112 +75,117 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
 /** Get the new list of stops from datastore onload */
 function getStopsOnload(){
   clearStops();
   fetch('/api/stop')
   .then(response => response.json())
   .then((stopsResponse) => {
-    stopsResponse.forEach((stop)=>{
-      stops.push(stop);
-    });
+    if(stopsResponse != null){
```
renau06, line 93: i think the if statement can me omitted becase if stopsResponse is null, the foreach loop will just not run. I am not completely positive on that though, but may want to check

byway/src/main/webapp/routepage.js
```diff
@@ -70,112 +75,117 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
-  btn.className =  "btn rec-btn";
-  btn.addEventListener("click", function() {
-    updateStops(stop, "remove");
+function createStopButton(stop){
+  const stopBtn = document.createElement('button');
+  stopBtn.innerText = stop;
+  stopBtn.className =  "btn rec-btn";
+  stopBtn.addEventListener("click", function() {
+    // remove place from stops and waypoints set
+    stops = new Set([...stops].filter(stopObj => stopObj!= stop));
```
Leo2452, line 123: nit: spacing towards the end when using the not equal operator

byway/src/main/webapp/routepage.js
j-vent: I got an error in the web dev console when the JSON response was null, so I just included that check.

byway/src/main/webapp/routepage.js
j-vent: Fixed.

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
+ *  @return {button} stopBtn a button showing a selected stop
  */
-function renderStop(stop){
-  const stopList = document.getElementById('stop-list');
-  let btn = document.createElement('button');
-  btn.innerText = stop;
-  btn.className =  "btn rec-btn";
-  btn.addEventListener("click", function() {
-    updateStops(stop, "remove");
+function createStopButton(stop){
```
landondare, line 142: Update function comment

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
+  const stopBtn = document.createElement('button');
+  stopBtn.innerText = stop;
+  stopBtn.className =  "btn rec-btn";
+  stopBtn.addEventListener("click", function() {
+    // remove place from stops and waypoints set
+    stops = new Set([...stops].filter(stopObj => stopObj != stop));
```
landondare, line 148: For the stops set, you could just delete the single stop instead instead of filtering the whole array. 
```javascript
stops.delete(stop);
```

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
+ *  @return {button} recBtn a button showing a recommended place
  */
-function renderRec(rec){
-  const recsList = document.getElementById('rec-list');
-  let btn = document.createElement('button');
-  btn.innerText = rec;
-  btn.className =  "btn rec-btn";
-  btn.addEventListener("click", function() {
-    updateStops(rec, "add");
+function createRecButton(rec){
```
landondare, line 199: Update function comment

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
+  const recBtn = document.createElement('button');
+  recBtn.innerText = rec;
+  recBtn.className =  "btn rec-btn";
+  recBtn.addEventListener("click", function() {
+    // TODO: add to recs later for better visuals on html
+    stops.add(rec);
+    //  if new place is add it to the selected stops and redraw route
```
landondare, line 206: nit: delete "is"

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
 // limitations under the License.
-
-
-// copies of recommendations and selected stops, stored as arrays for synchronous updating
-let recs = [];
-let stops = []; // TODO: Make set to avoid duplicates
-
+ 
+// copies of recommendations and selected stops, stored as sets for synchronous updating
+let recs = new Set;
```
landondare, line 16: Missing parenthesis: `let recs = new Set()`. Update this and other set variables.

byway/src/main/webapp/routepage.js
```diff
@@ -11,171 +11,206 @@
...
+  for (i = 0; i < route.legs.length; i++) {
+    totalDist += route.legs[i].distance.value;
+    totalTime += route.legs[i].duration.value;
+  }
+
+  totalDist = (totalDist / 1000).toFixed(2);
+  hours = Math.floor(totalTime / 3600);
+  minutes = Math.round((totalTime - hours*3600) / 60);
+ 
+  document.getElementById("distance").innerText = totalDist + "km";
```
landondare, line 99: How are the units of distance determined? Does it default to meters?

j-vent:
> Seems like your branch may be out of sync, unless you intend to delete these java servlet files.

Yup that was intentional because the waypoint selection stuff is all js and I didn't want the servlet code to be reviewed in this PR and my other open PR.

byway/src/main/webapp/routepage.js
j-vent: Could you please explain what you mean? One of them is `@param` and the other is `@return`.

byway/src/main/webapp/routepage.js
j-vent: Oh yeah you're absolutely right that's a funny error, changed.

byway/src/main/webapp/routepage.js
j-vent: done.

byway/src/main/webapp/routepage.js
j-vent: I saw it instantiated as just Set in some other code and didn't question it. My bad, changed it to the proper way.

byway/src/main/webapp/routepage.js
j-vent: It says "km" in the text which I thought was sufficient, but I will comment for clarity since it does start as meters.

landondare: This was more of a question about the API rather than a comment on your code. Just wondering if this is specified in your request or if it always returns meters. May be useful to toggle between kilometers and miles in the future if possible, but don't worry about it now.

byway/src/main/webapp/routepage.js
landondare: The function title says render stop list, but this now creating a button for the stop list instead. It doesn't actually add it to the stop list.

byway/src/main/webapp/routepage.js
j-vent: done.

byway/src/main/webapp/routepage.js
j-vent: I added a comment saying the API defaults to meters.

byway/pom.xml
```diff
@@ -43,13 +43,6 @@
...
 
-    <!-- https://mvnrepository.com/artifact/com.google.flogger/flogger -->
-    <dependency>
-      <groupId>com.google.flogger</groupId>
-      <artifactId>flogger</artifactId>
-      <version>0.5.1</version>
-    </dependency>
-
     <!-- https://mvnrepository.com/artifact/com.google.flogger/flogger -->
     <dependency>
```
zstewar1, line 47: Please check this file out back to the version in master.

byway/src/main/java/com/google/sps/data/Recommendation.java
```diff
@@ -23,17 +23,20 @@
...
     this.placename = placename;
+   
   }
 
   public static Recommendation fromEntity(Entity entity){
     checkArgument(entity.getKind().equals(KIND), "Wrong Entity kind, expected %s, got %s", KIND, entity.getKind());
     long id = entity.getKey().getId();
     String placename = (String) entity.getProperty("placename");
+
     return new Recommendation(id, placename);
```
zstewar1, line 40: Thought you deleted this class? Do you need to merge master to pick up that change?

byway/pom.xml
j-vent: I checked out pom.xml and the java files from master.

byway/src/main/java/com/google/sps/data/Recommendation.java
j-vent: After checkout I deleted Recommendation.java and Stop.java since they are both in master.

byway/src/main/java/com/google/sps/servlets/RecsServlet.java~HEAD
```diff
@@ -0,0 +1,71 @@
...
+    List<Recommendation> recs = new ArrayList<>();
+
+    for (Entity entity: results.asIterable()){
+      recs.add(Recommendation.fromEntity(entity));
+    }
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(recs));
+  }
+  
+}
```
zstewar1, line 71: Looks like you accidentally comitted a new file called RecsServlet.java-HEAD. Delete this?

byway/src/main/webapp/routepage.js
```diff
@@ -56,42 +70,75 @@ function initMap() {
...
  */
-function calcRoute(directionsService, directionsRenderer, start, end) {
+
+function calcRoute() {
   let request = {
     origin:  start,
     destination: end,
-    travelMode: 'DRIVING'
+    travelMode: 'DRIVING',
+    waypoints: Array.from(waypoints),
```
zstewar1, line 79: Is waypoints supposed to be ordered? Sets generally don't guarantee any kind of ordering.

byway/src/main/webapp/routepage.js
```diff
@@ -56,42 +70,75 @@ function initMap() {
...
+  }
+
+  totalDist = (totalDist / 1000).toFixed(2);
+  let hours = Math.floor(totalTime / 3600);
+  let minutes = Math.round((totalTime - hours*3600) / 60);
+ 
+  document.getElementById("distance").innerText = totalDist + "km";
+  document.getElementById("duration").innerText = 
+    (hours == 0) ? minutes + " mins" : hours + " hr " + minutes + " mins";
+}
```
zstewar1, line 115: Instead of writing these directly to DOM, consider separating calculation from display by having this method return a plain object with the computed values, and then writing that to dom as a separate operation.

byway/src/main/webapp/routepage.js
```diff
@@ -12,40 +12,54 @@
...
-let recs = [];
-let stops = []; // TODO: Make set to avoid duplicates
-
+ 
+// copies of recommendations and selected stops, stored as sets for synchronous updating
+let recs = new Set();
+let stops = new Set();
+ 
+// location representations of stops array
+let waypoints = new Set();
```
zstewar1, line 21: Try to avoid storing duplicate data. It's best to have a 'single source of truth' and just transform or extract relevant parts of that data as needed, rather than trying to maintain independent copies of the data with different version so f the same information. If there's extra information about the stops, just store it in the form of waypoints, and when you just want the 'stops' values, extract just those. If there's no extra information and `waypoints` is just a different representation of the same information, just save `stops` and compute waypoints as needed. For example, I notice that whenever you use `waypoints` you use `Array.from` anyway, so it's not like you're avoiding constructing a new array every time by doing this.

`Array.from(stops).map(stop => ({location: stop}))` would construct a pretty-much equivalent array without having to store duplicate data.

byway/src/main/webapp/routepage.js
```diff
@@ -12,40 +12,54 @@
...
+ 
+// object that communicates with the GMaps API service
+let directionsService;
+ 
+// object that renders display results on the map
+let directionsRenderer;
+ 
+// TODO: get from Trip key
+let start;
+let end;
```
zstewar1, line 31: Are start and end also part of `stops` or no?

byway/src/main/java/com/google/sps/servlets/RecsServlet.java~HEAD
j-vent: Done.

byway/src/main/webapp/routepage.js
j-vent: Yeah, I made them sets so duplicate ones don't get added but I think I need ordering now that I want to use the optimize feature.

byway/src/main/webapp/routepage.js
j-vent: Thanks for reminding me of this method, I had a hard time Googling how to do something like this. Will change.

byway/src/main/webapp/routepage.js
j-vent: I will take it out of the stops bc it will map the place as both a waypoint and as start point. Also start == end for the roadtrip so I think I will include other major cities as waypoints.

byway/src/main/webapp/routepage.js
j-vent: Okay, I've separated them into two functions. Please let me know if you had something else in mind.

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -74,6 +74,15 @@ public Trip(
     return Collections.unmodifiableList(this.route);
   }
 
+  /**
+   * Sets the route of the trip with a list of stops and destinations as plain text.
+   *
+   * @param route list of user-selected stops and destinations
+   */
+  public void setRoute(Collection<String> route) {
+    this.route = new ArrayList<>(route);
```
zstewar1, line 83: keep `route` as final and  use `this.route.clear(); this.route.addAll(route);` Using clear + assign here is nice because it potentially avoids allocating a new array, and lets you keep the field as final.

byway/src/main/webapp/routepage.js
```diff
@@ -69,60 +100,115 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+ */
+function computeDistanceTime(result) {
+  let totalDist = 0;
+  let totalTime = 0;
+  // full route
+  let route_response = result.routes[0];
+  let waypoint_order = route_response.waypoint_order;
+  
+  let route_copy = [...route];
+  for(let i = 0; i < route.length; i++){
```
landondare, line 118: This snippet of code seems to be unrelated to the logic for computing the distance and time. It also doesn't look like the updated route data is used by the method. Please move this to its own section and document its purpose.

byway/src/main/webapp/routepage.js
```diff
@@ -12,55 +12,86 @@
...
-  document.getElementById("route").addEventListener("click", function() {
-    calcRoute(directionsService, directionsRenderer, start, end);
+  
+}
+
+/** Displays route containing waypoints overtop the map. */
+function calcRoute() {
+  let request = {
+    origin:  start,
+    destination: end,
```
zstewar1, line 65: Does having these as explicit origin/destination not prevent optimizeWaypoints from moving them into the middle of the route?

byway/src/main/webapp/routepage.js
```diff
@@ -69,60 +100,115 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
+function computeDistanceTime(result) {
+  let totalDist = 0;
+  let totalTime = 0;
+  // full route
+  let route_response = result.routes[0];
+  let waypoint_order = route_response.waypoint_order;
+  
+  let route_copy = [...route];
+  for(let i = 0; i < route.length; i++){
+    route[i] = route_copy[waypoint_order[i]];
```
landondare, line 119: What is this doing exactly? Is it re-ordering the route list based on the waypoint order?

Additionally, is the `route` array always guaranteed to be the same size as the `waypoint_order`? Doesn't this function get called when a user removes a waypoint? Please add some additional documentation for this section.

byway/src/main/webapp/routepage.js
```diff
@@ -69,60 +100,115 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
-    if(stopsResponse != null){
-      stops.push(...stopsResponse);
+  .then((trip) => {
+    if(trip != null){
+      start = end = trip.start;
+      destinations = trip.destinations;
+      route.push(...trip.route)
+
+      calcRoute();
+      renderRouteList();
```
landondare, line 168: Note that `calcRoute()` will call `updateRoute()` will call `renderRouteList()`, making this call redundant. This can be removed.

byway/src/main/webapp/routepage.js
zstewar1: When doing a pure computation, it's best to avoid "side effects", that is modifying something other than the data you're working on  (or modifying the data you're given at all!). In this case, you have a function which could compute the length of the trip without causing any other effects, except for that you call the function to update the DOM at the bottom. A function which has no side-effects is more generally useful than one with side-effects, because a function without side-effects can be used from anywhere without worrying about what it might do to the page. So like, if you decide you need the length of the trip for some other computation, you could call the function without worrying that it also updates the page. This is why I say to return the computed value as a plain object, like `return {totalDist, hours, minutes};` rather than directly calling into code that updates the DOM. Then whever called the functio ncan decide what to do with that data, wheter that's update the DOM or use it as part of a different calculation.

byway/src/main/webapp/routepage.js
```diff
@@ -12,55 +12,86 @@
...
+// object that renders display results on the map
+let directionsRenderer;
+ 
+// TODO: get from Trip key
+let start = "";
+let end = "";
+ 
 if (document.readyState === 'loading') {  // Loading hasn't finished yet
   document.addEventListener('DOMContentLoaded', loadData);
+  console.log("loading");
```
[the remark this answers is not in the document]
j-vent: I already removed these in a prior commit, maybe at the same time as you commented.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -42,42 +41,46 @@
...
+  // TODO: get key from query string params
   private final Key key = KeyFactory.createKey(Trip.DATASTORE_ENTITY_KIND, 1234);
 
-  /* Passes saved destinations stops (if any) to be shown in the schedule panel */
+  /* Passes saved route to be shown in the schedule panel */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<String> stops = Collections.emptyList();
-    Entity entity;
+    List<String> destinations = Collections.emptyList();
```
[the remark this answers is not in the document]
j-vent: Already removed these variables in a prior commit.

byway/src/main/webapp/routepage.js
```diff
@@ -69,60 +100,115 @@ function calcRoute(directionsService, directionsRenderer, start, end) {
...
-  stopBtn.addEventListener("click", function() {
-    // TODO: delete from recs later for better visuals on html
-    stops = stops.filter(function(stopObj){
-      return stopObj != stop;
-    })
-    updateStops();
+function createRouteButton(waypoint){
+  const routeBtn = document.createElement('button');
+  routeBtn.innerText = waypoint;
+  routeBtn.className =  "btn rec-btn";
```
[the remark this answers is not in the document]
j-vent: Oh yes I was thinking of doing that in another PR w all the style changes, but I can add it here since it's not too much work.

byway/src/main/webapp/routepage.js
```diff
@@ -12,55 +12,86 @@
...
+  directionsService.route(request, function(response, status) {
+    if (status == 'OK') {
+      directionsRenderer.setDirections(response);
+      computeDistanceTime(response);
+    } else {
+      window.alert("Could not calculate route due to: " + status);
+    }
   });
+
+  updateRoute();
```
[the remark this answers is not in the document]
j-vent: I don't think the order is too important for these two, but I'll keep this in mind for other functions where order is important. Changed.

byway/src/main/webapp/routepage.js
j-vent: Will do, I just did it out of convenience since the directions response was already passed in as a param, but it is bad practice.

byway/src/main/webapp/routepage.js
j-vent: I moved it out to another function as per your comment above and added documentation. Sorry about that, I forgot to comment that section altogether.

byway/src/main/webapp/routepage.js
j-vent: Oh good catch, thanks! Changed.

byway/src/main/java/com/google/sps/data/Trip.java
j-vent: Changed. Thanks for explaining `final` in mutable classes.

byway/src/main/webapp/routepage.js
j-vent: Yes it does, but I removed start/end from the route list because if they are they will become waypoints. I don't think having generateRoute perform another request is necessary, will remove.

byway/src/main/webapp/routepage.js
j-vent: Ohh I see what you mean. I changed my implementation to this.

byway/src/main/webapp/routepage.js
```diff
@@ -12,117 +12,198 @@
...
-  const stopList = document.getElementById('stop-list');
-  if(stopList != null){
-    stopList.innerText = "";
+/** Add the starting location back to the schedule panel
+ *  TODO: Delete markers for recommended stops not selected.
+ *  TODO: Disable usage after? Don't want to keep adding to list. 
+ */
+function generateRoute() {
+  route.splice(0,0,start);
+  route.splice(route.length, 0, end);
```
zstewar1, line 88: What is this doing? Try to keep the state of the `route` list consistent, that is, make it so it either always contains the start and end, or never contains the start and end. Having a function which adds the start and end without clear tracking of whether the start and end are currently included in the list is error prone, because it becomes very easy to run this add operation extra times even when the list already contains the start/end.

j-vent: I just wanted to add back the start/end location to the schedule panel when the user has finalized the stops they want to visit,  and the way I did this was add them back to the route. It's more just a visual thing. Yeah, I didn't add anything to prevent this action from being taken multiple times. Do you have any suggestions as to how to improve this?

byway/src/main/webapp/routepage.js
```diff
@@ -12,117 +12,198 @@
...
+  let route_response = response.routes[0];
+
+  for (let i = 0; i < route_response.legs.length; i++) {
+    // in meters
+    totalDist += route_response.legs[i].distance.value;
+    // in seconds
+    totalTime += route_response.legs[i].duration.value;
+  }
+  
+  let distance = (totalDist / 1000).toFixed(2);
```
landondare, line 125: Nit: add comment about units for distance here and in function comment.

byway/src/main/webapp/routepage.js
```diff
@@ -12,117 +12,198 @@
...
+  let hours = Math.floor(totalTime / 3600);
+  let minutes = Math.round((totalTime - hours*3600) / 60);
+
+  return {distance, hours, minutes};
```
landondare, line 129: Interesting, I've never seen this shorthand in javascript before. It essentially creates a dictionary where the keys are the variable names. Neat!

You may want to update you documentation for the return type slightly to make accessing the returned value more clear. 

```javascript
@return {'distance' : number, 'hours': number, 'minutes': number}
```

byway/src/main/webapp/routepage.js
```diff
@@ -12,117 +12,198 @@
...
+  .then((trip) => {
+    if(trip != null){
+      start = end = trip.start;
+      destinations = trip.destinations;
+      route.push(...trip.route)
+
+      calcRoute();
+    }
+    else{
+      console.log("Could not retrieve any routes nor destinations associated with this trip. Please reload page and try again.");
```
landondare, line 165: Nit: split this into multiple lines so that it is less than 100 characters long

byway/src/main/webapp/routepage.js
j-vent: I have the units written in the for loop above this but I'll move it for readability, though I did list them in the function comments in the `@return`.

byway/src/main/webapp/routepage.js
zstewar1: Just add the start/end when creating elements on the page, without changing the underlying data stored in the page-wide variables.

byway/src/main/webapp/routepage.js
j-vent: Okay, will update that. Your formatting is more readable.

byway/src/main/webapp/routepage.js
j-vent: Changed.

byway/src/main/webapp/routepage.js
j-vent: Oh I see, I changed generateRoute() to reflect this.

PR #24, opened 2020-08-11
Partition route

Leo2452:
For any given route, retrieve the steps of the path and add the midpoint lat/lng coords of the step to an array if it is long enough (limited by a MIN_DISTANCE of 4000 (meters) as of now). Use said array to search for user interests around that array to load recommendations, currently showing as markers.

Uses hard-coded interests and a simple route, point A to point B.


byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+let map;
+let placesService;
+
+// Measured in meters
+let RADIUS = 1000;
+let MIN_DISTANCE = 4000;
+
+let interests = ["park"];
+let destinations = new Set();
+let regions = [];
```
j-vent, line 33: May be good to comment what exactly regions holds as it is not intuitive.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Add start and end points of the route to regions to search around.
+ * Temporary set up to mimic adding several endpoints between legs.
+ */
+function addDestinations(start, end) {
+  destinations.add(start);
+  destinations.add(end);
+  for(destination of destinations) {
+    regions.push(destination);
```
j-vent, line 66: Can use the spread operator instead: ``` regions.push(...destinations) ```. More example usage can be found at: https://www.javascripttutorial.net/es6/javascript-spread/

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+  const myRoute = directionResult.routes[0].legs[0];
+  for(step of myRoute.steps) {
+    let avgLat = (step.start_location.lat() + step.end_location.lat()) / 2;
+    let avgLng = (step.start_location.lng() + step.end_location.lng()) / 2;
+    let avgLoc = {lat: avgLat, lng: avgLng};
+    if(step.distance.value > MIN_DISTANCE) {
+      regions.push(avgLoc);
+    }
+  }
+  console.log("size of regions: " + regions.length);
```
j-vent, line 113: Remove debugging comment.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/**
+ * Places markers on the locations found from textSearch.
+ * Temporarily limit the amount of suggestions.
+ * TODO: Store more results and limit on UI with option to "show more"
+ * @param {PlaceResults[] results} places found with PlaceResult type.
+ */
+function addRecommendations(results) {
+  let maxRecommendations = 1;
```
j-vent, line 161: Use `const MAX_RECOMMENDATIONS` instead.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+ * Places a marker onto the map at the specified location.
+ * @param {PlaceResult place} Object containing information
+ * about a place
+ */
+function placeMarker(place) {
+  let infoWindow = new google.maps.InfoWindow({
+    content: place.name
+  });
+  const image = {
+    url: './images/stopsMarker.png',
```
j-vent, line 183: Use an absolute path 'images/...' instead of a relative one.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,195 @@
...
+    scaledSize: new google.maps.Size(40,40)
+  };
+  let marker = new google.maps.Marker({
+    position: place.geometry.location,
+    map: map,
```
j-vent, line 188: I believe if the key and variable are the same, you can use a Javascript shorthand where you would just list `map`.

byway/src/main/webapp/recommendations.js
Leo2452: Updated

byway/src/main/webapp/recommendations.js
Leo2452: Updated

byway/src/main/webapp/recommendations.js

byway/src/main/webapp/recommendations.js

byway/src/main/webapp/recommendations.js
Leo2452: Fixed, included root directory `/images/...`

byway/src/main/webapp/recommendations.js
Leo2452: Updated

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,194 @@
...
+
+// Contains google.maps.LatLng objects.
+// Used as a center point to search around a region.
+let regions = [];
+
+/**
+ * Initializes the webpage with a map and other google
+ * services. Creates a route between two endpoints.
+ */
+function initialize() {
```
renau06, line 42: maybe be more specific about what this function is initializing

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,194 @@
...
+  // Modified version of Justine's implementation
+  var directionsService = new google.maps.DirectionsService();
```
renau06, line 44: i think zach/landon had mentioned before to try not to use var, and use let or const instead. not sure if it is different when using the api but maybe try to let/const  to that for these variables

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,194 @@
...
+ * Creates a route between two points and loads onto the map.
+ * Finds points along the path to load the regions Array.
+ * Loads recommendations centered around these points.
+ * @param {DirectionsService directionsService} finds directions
+ * @param {DirectionsRenderer directionsRenderer} renders the route
+ * @param {LatLng start} starting point location
+ * @param {LatLng end} ending point location
+ */
+function calcRoute(directionsService, directionsRenderer, start, end) {
+  var request = {
```
renau06, line 81: same here, use let or const

byway/src/main/webapp/recommendations.js
Leo2452: Updated for all `var` types, also updated some `let` types

byway/src/main/webapp/recommendations.js
Leo2452: Updated

byway/src/main/webapp/recommendations.js
Leo2452: Updated/described temp set up

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,197 @@
...
+        query: interest
+      }
+      const results = await new Promise(resolve => {
+        placesService.textSearch(request, (result, status) => {
+          if(status == "OK") {
+            resolve(result);
+          } else if(status == "OVER_QUERY_LIMIT") {
+            alert("Showing limited results");
+          } else {
+            alert("Status: " + status);
```
zstewar1, line 148: Note: if you never use `resolve` the promise will never complete, and the outer function will hang forever on the `await`. This means if you call this function more than once and have it hang, it will leak memory, as an increasing number of incomplete promises will be left sitting around waiting to finish, even though they never will. You can deal with this by taking not just `resolve` but also `reject`, so you can send an error to the `await`. Even if you don't handle that error, you at least cause the `await` block to throw, which aborts the function and doesn't leak memory.

```js
new Promise((resolve, reject) => {
// ...
if (status === "OK") {
  resolve(result);
} else {
  reject(new Error(status)); // or similar. What is "result" anyway when status != OK?
}
// ...
});

File: byway/src/main/webapp/generator.html [hunk 11]
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>My Portfolio</title>
```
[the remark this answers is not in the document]
Leo2452: Updated, along with description to indicate this temporary page will eventually be removed and the js will reference the original route page instead.

byway/src/main/webapp/generator.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,204 @@
...
+  directionsRenderer.setMap(map);
+  document.getElementById("route").addEventListener("click", () => {
+    calcRoute(directionsService, directionsRenderer, start, end);
+  });
+  placesService = new google.maps.places.PlacesService(map);
+}
+
+/**
+ * Add start and end points of the route to regions to search around.
+ * Temporary set up to mimic adding several endpoints between legs.
```
landondare, line 66: Document  function parameters

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,204 @@
...
+          }
+        });
+      });
+      if(results !== null) {
+        addRecommendations(results);
+      }
+    }
+  }
+  if(isLimited) {
+    alert("Showing limited results");
```
landondare, line 160: Is this message for you the programmer or for an actual user? If it is just helpful for you, it would probably be better to just log this instead of interrupting the page with an alert. If it is a message meant for the user, then it would be better to display a message in the HTML rather than an alert. A small paragraph in red for example to explain the incomplete results.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * Go through a user's interests and search for places
+ * fitting those interests. Search around the regions
+ * previously found using textSearch.
+ */
+async function loadRecommendations() {
+  let isLimited = false;
+  for(interest of interests) {
+    for(region of regions) {
+      await delayPromise(250);
```
landondare, line 136: What's the purpose of this delay? Is it to prevent too many requests going to the maps API at the same time? Can you please add some comments explaining the logic here.

Leo2452: Updated

byway/src/main/webapp/recommendations.js
Leo2452: This was meant for the user. I fixed this, but the issue stands when there are multiple "OVER_QUERY_LIMITS" or other status codes s.t. `status !== "OK"`. I'm not sure how to await this part to display the message once, and only after the results are finished.

byway/src/main/webapp/recommendations.js
Leo2452: Updated

byway/src/main/webapp/recommendations.js
Leo2452: I updated to catch the exception as well and alert the user of the message. The result is null when `status !== OK`, but it seems after awaiting, the newly-named `placesFound` (previously named `results`) is undefined rather than null. I also added a check to call addRecommendations only when it is defined to avoid issues.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/**
+ * Saves the places generated from interest and location in sessionStorage as key/value pairs.
+ * @param {TextSearchRequest} request with unique location and interest
+ * @param {PlaceResults[]} placesLoaded List of places related to interest
+ */
+function savePlacesFromInterests(request, placesLoaded) {
+  sessionStorage.setItem(JSON.stringify(request), JSON.stringify(placesLoaded));
+}
```
Leo2452, line 231: Note: When I debugged with the console, it seems like part of how the sessionStorage sets items is not entirely reliable. I got a warning that `open_now()` is deprecated as of Nov. 2019 and will be removed Nov. 2020, and to instead use `is_open()`. I tried to see if there was an update to sessionStorage / alternative, but could not find anything on this.

Leo2452: The issue seems to be storing the placesLoaded containing `placeResults` in the sessionStorage. Since it is storing the entire place object, it contains a field, `PlaceResult.opening_hours.open_now`, which is deprecated/willl be replaced. Since we aren't storing this in datastore and instead the current browser session, I think this should be fine?

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+        radius: RADIUS,
+        query: interest
+      }
+      const recommendationsSaved = sessionStorage.getItem(JSON.stringify(request));
+      if(recommendationsSaved !== null) {
+        addRecommendations(request, JSON.parse(recommendationsSaved));
+      } else{
+        // Prevent hitting the query limit from the maps API
+        // by setting a timeout in delayPromise
+        await delayPromise(250);
```
landondare, line 153: Does this actually work the way you intend? It looks like it is just setting a timer for 250 ms before continuing execution. Is the idea wait for multiple calls to `loadRecommendations()`

byway/src/main/webapp/recommendations.js
zstewar1: Note that using `stringify` on objects that are part of a class may have odd/unexpected results, especially when converting back from a string (it will end up as a plain JS object, not an instance of the original class). When storing things as JSON you may need/want to convert them to plain objects ("records") containing only the fields you care about.

byway/src/main/webapp/recommendations.js
Leo2452: This timeout is an intermediate step when making a new textSearch request. When doing `await findPlacesWithTextSearch()`, it waits for the function to return a new promise guaranteed to be completed before it continues executing. Then when it finishes adding them, it continues in the for loop to find the new results for a different interest/region.

If I remove this `await delayPromise()` line, the code awaits like normal when making a new call to `findPlacesWithTextSearch()`. The issue I saw, however, was when refreshing, as the await seems to not "wait" as much with results that were already generated. Not sure how this works, I assumed it was cached or something similar. But, it hit the query limit if too many results were found before refreshing, since loading them after the refresh didnt work the same way for previously-loaded entries.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * Provides a message to the user if there was an issue
+ * with their request.
+ * TODO: Make a class to indicate as a "warning" and make text red.
+ * Also, organize on front end.
+ * @param {String} msgFromService status String from placesService
+ */
+function alertUser(msgFromService) {
+  let messageContainer = document.getElementById("message-container");
+  let message = document.createElement('p');
```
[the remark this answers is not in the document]
Leo2452: Updated. I might try to find a better way to load the message on screen per unique status message, without having too add multiple "OVER_QUERY_LIMIT" messages or any other status. My update might still have this problem, since it is listing the statuses and there's no check if one already exists.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+} else {
+  initServices();
+}
+
+let map;
+let placesService;
+
+// Measured in meters
+const RADIUS = 1000;
+const MIN_DISTANCE = 4000;
```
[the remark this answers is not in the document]
Leo2452: Updated

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+  savePlacesFromInterests(request, placesLoaded)
+}
+
+/**
+ * Saves the places generated from interest and location in sessionStorage as key/value pairs.
+ * @param {TextSearchRequest} request with unique location and interest
+ * @param {PlaceResults[]} placesLoaded List of places related to interest
+ */
+function savePlacesFromInterests(request, placesLoaded) {
+  sessionStorage.setItem(JSON.stringify(request), JSON.stringify(placesLoaded));
```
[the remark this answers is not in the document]
Leo2452: I had thought the issue was only with refreshing, but I checked and the pre-loaded results after awaiting the promise hits the query limit in a new tab as well. The concern with using localStorage is having to clear the information after closing the tabs. ie if theres two tabs on the routepage, and i clear when a tab is closed, it would presumably affect the other tab still open. 

Another point is this localStorage would act like the datastore from the backend, except it would be saved on the server and not accessible on the front end via dev tools. The point of avoiding datastore was to save recommendations was to avoid saving/deleting unselected recommendations. The reason i saved on the front end temporarily was to await hitting the query limits on refresh (and new tabs), but there seem to be caveats to both approaches. 

I'm not entirely sure what the best course of action is.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,254 @@
...
+    placesService.textSearch(request, (result, status) => {
+      if(status === "OK") {
+        resolve(result);
+      } else {
+        reject(new Error(status));
+      }
+    });
+  }).catch((error) => {
+    alertUser(error.message);
+  });
```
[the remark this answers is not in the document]
Leo2452: I resolved the result, even if the status was not ok, and checked afterwards if the result was null rather than undefined through the catch chaining. I'm not sure if it is what the resolve was intended to do, ie if one should only resolve when it is successful, but checking the result for null-ness was what I wanted to have to indicate a failed request.

Updated

byway/src/main/webapp/recommendations.js
zstewar1: @landondare have a look at the constructor for `Promise` works.

Leo2452:
Using localStorage would mean the data persists throughout the tabs opened, but there is no good way to know when to clear it. I tried to clear when the tab is closed, but then that acts like sesisonStorage. I couldn't find a way to clear only when all browser tabs were closed, especially since there could be two unique users using at a time.

I am using sessionStorage to preserve findings in the current tab, and if they open a new tab, `delayPromise` sets a timeout and loads it to the sessionStorage as it loads more results. I couldn't find a workaround, but this method seems to be effective and avoids weird edge cases to handle with localStorage (i.e. opening 3 tabs and closing one tab while results are still loading, loading results on two windows simultaneously, etc)

byway/src/main/webapp/recommendations.js
Leo2452: The objects being stringified are requests (key) and PlaceResults[] (value). So far, the properties we have been using are accessed via fields (`.geometry.location`, `.name`, etc) and the plain JS object seems to accomplish this. We can use a LatLng literal object in place of the LatLng custom object from the API, but we can't access functions the same way with the plain JS object. I don't think we are using methods from the classes, but I will keep this in mind/relay with my teammates to be wary of this.

Or rather, do either of you recommend storing this information on the client side a different way? This way works and we aren't using class-specific methods that are lost in JSON translation, but if there's another recommendation I could look more into it.

byway/src/main/webapp/recommendations.js
```diff
@@ -0,0 +1,257 @@
...
+ * @param {TextSearchRequest} request object with location, radius and query fields.
+ * @return promise either with PlaceResult[] results or undefined if rejected.
+ */
+function findPlacesWithTextSearch(request) {
+  return new Promise((resolve, reject) => {
+    placesService.textSearch(request, (result, status) => {
+      if(status === "OK") {
+        resolve(result);
+      } else {
+        alertUser(status);
```
landondare, line 177: You'll want to do the opposite in the case that it was successful. You'll probably want to reset the alerts in general every time you call this method. This way you don' show any errors that are no longer relevant. you could reset the visibility of the alerts before checking the status, and then only the newest, if any, errors wil be displayed

```javascript
function resetUserAlerts() {
    document.getElementById("message-container")style.visibility = 'hidden';
    document.getElementById("query-limit").style.visibility = 'visible';
    document.getElementById("incomplete-results").style.visibility = 'visible';
    document.getElementById("other-statuses").style.visibility = 'visible';
}
```

Leo2452: This function is called inside a loop for multiple requests at a time. I want to alert the user if any request is not successfully filled, ie if one has zero results or it hits a query limit in the middle of the requests. If I add this code, the user may not get alerted if a status was not OK. Instead, I want a message to show up if at least one status other than OK was hit, and to list off all statuses hit if more than one.

I modified the code to simple show a general message, "showing limited results," and having it show the status returned from the PlacesService. I added a check to not redundantly show if a query limit was hit 10 times via sets.

landondare: Thats a good catch with the loop. You wouldn't want to remove the alert if one of the iterations is successful. However, you may still want to hide the alerts if the user calculates the route multiple times. No reason to show the error if the second attempt succeeds. You could reset the alerts in the `loadRecommendations() ` functions instead, so that it would be outside this loop.

Leo2452: Good point, forgot to account for that. Fixed

PR #25, opened 2020-08-12
Dataschema helper

j-vent:
Added toEntity() function which converts UserInfo object to a Datastore object, as well as a "setter" function addTripId() to add a new trip ID to the list.


j-vent:
No longer needed. Added directly to another PR.

PR #27, opened 2020-08-17
Style interests page

Leo2452:
Style the interests page to include: progress bar, header, next button, etc. Also, apply color scheme and other stylistic changed.

Also, I changed tripId -> tripKey throughout files where I used tripId or referenced the trip as id. This change did not affect functionality, but it was done for consistency and proper access in datastore.

### DIFF
--- a/byway/src/main/webapp/base.js
+++ b/byway/src/main/webapp/base.js
@@ -1,4 +1,4 @@
-/* exported setProgressBar, getTripKeyFromUrl, configureTripKeyForNextPage */
+/* exported configureTripKeyForPath, getTripKeyFromUrl, setProgressBar */
 /** 
 * Sets Progress Bar to correct location based on the page number
 * @param {int} pageNumber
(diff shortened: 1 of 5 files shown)


byway/src/main/webapp/base.js
```diff
@@ -12,30 +12,30 @@ function setProgressBar(pageNumber){
 }
 
 /**
- * Go through url to retrieve the trip id.
- * @returns String of trip ID
+ * Go through url to retrieve the trip key.
+ * @returns String of trip key
```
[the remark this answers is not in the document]

byway/src/main/webapp/track-interests.js
```diff
@@ -83,22 +84,22 @@ function createButtonForPlace(place) {
...
  * may be any type that is convertible to an array via Array.from. It will
  * be sent to the server as JSON in the post body.
- * @param {String} tripId value for tripId
+ * @param {String} tripKey value for tripKey
  * @param {Array} [userInterests] interests selected by user
  */
-function fetchPlaces(tripId, /* optional */ userInterests) {
-  const url = '/api/places?' + new URLSearchParams({tripId}).toString();
+function fetchPlaces(tripKey, /* optional */ userInterests) {
+  const url = '/api/places?' + new URLSearchParams({tripKey}).toString();
```
renau06, line 102: could you use the configureTripKeyForNextPage function for this line too?

byway/src/main/webapp/base.js
```diff
@@ -12,30 +12,30 @@ function setProgressBar(pageNumber){
...
-function configureTripIdForNextPage(tripId, nextPagePath) {
-  if(tripId === null) {
+function configureTripKeyForNextPage(tripKey, nextPagePath) {
+  if(tripKey === null) {
     // Send back to page where a trip can be made
-    alert("tripId not created! Create a new trip.");
+    alert("tripKey not created! Create a new trip.");
     window.location.href = '/index.html';
   } else {
     let nextPage = document.getElementById("next-button");
```
renau06, line 38: maybe you shouldn't include the let nextPage=document.getElementbyId("next-button") if we are all going to be using this function since we might name it something different. I would recommend just returning url

byway/src/main/webapp/track-interests.js
Leo2452: Good point, updated the function

byway/src/main/webapp/base.js
Leo2452: Updated to be more general and fill any path with the tripKey param

PR #30, opened 2020-08-18
Changeto store place ids

renau06:
Change destinations.js and destinationsservelt.java to store place ids instead of place names


byway/src/main/webapp/destinations.js
```diff
@@ -224,15 +236,52 @@ window.onload = function(){
...
+    query: String(document.getElementById("destinations-search-box").value),
+    fields: ["place_id"]
+  };
+  service.findPlaceFromQuery(destinationRequest, (results, status) => {
+    if (status === google.maps.places.PlacesServiceStatus.OK) {
+      formData.append("destinations", results[0].place_id);
+      fetch(buildUrlWithParams("/api/destinations", {tripKey}), {method: 'POST', body:formData}).then((response)=>
+        response.json()).then(locationData => {
+        updateLocations(locationData);
+      });
```
Leo2452, line 278: nit: align and format code by having the fetch in one line, and separating following promises per line, ie

```
fetch(...)
.then(...)
.then(...)
```

byway/src/main/webapp/destinations.js
```diff
@@ -224,15 +236,52 @@ window.onload = function(){
...
   document.getElementById('user-input-form').addEventListener('submit', (event) => {
     event.preventDefault();
-    const formData = new  FormData(document.getElementById("user-input-form"));
-    fetch(buildUrlWithParams("/api/destinations", {tripKey}), {method: 'POST', body:formData}).then((response)=>
+    getDestinationPlaceId();
+    getStartPlaceId();
+  });
+}
+
+function getStartPlaceId(){
```
Leo2452, line 244: This function is very similar to `getDestinationPlaceId()`. I suggest making a function like `getPlaceId()` or something, that takes in some params to distinguish between destination or start (ie boolean or string literal). The only difference is what you pass into the formData ("start-location" vs "destinations") and calling either updateStartDestination or updateLocations. Checking these two places within one single function prevents having two separate functions that serve nearly the same purpose

byway/src/main/webapp/destinations.js
```diff
@@ -180,11 +180,23 @@ function addLocationToDom(results,container){
...
+    if (locationData.start == "" || locationData.start == null){
       document.getElementById('start-search-box').value = "";
     }
     else{
-      document.getElementById('start-search-box').value = locationData.start;
+      const request = {
+        placeId: locationData.start,
+        fields: ["formatted_address"]
+      };
+      let service = new google.maps.places.PlacesService(map);
```
Leo2452, line 191: Rather than declare multiple times, make into a global variable so it is made once and persists for all functions

byway/src/main/webapp/destinations.js
```diff
@@ -180,11 +180,23 @@ function addLocationToDom(results,container){
 * fills Start location Searchbox with previously input
 */
 function updateStartDestination(locationData){
```
Leo2452, line 182: Update comments within code throughout PR, along with including params passed in

byway/src/main/webapp/destinations.js
```diff
@@ -134,16 +134,16 @@ function addMarker(searchBox){
...
       };
       let service = new google.maps.places.PlacesService(map);
-      service.findPlaceFromQuery(request, (results, status) => {
+      service.getDetails(request, (place, status) => {
         if (status === google.maps.places.PlacesServiceStatus.OK) {
-          addLocationToDom(results, container);
+          addLocationToDom(place, container);
         }
         else{
           alert("Location Invalid");
```
j-vent, line 149: It's good to return the status code here too. Fix in other places also.

byway/src/main/webapp/destinations.js
```diff
@@ -224,15 +236,52 @@ window.onload = function(){
...
+      fetch(buildUrlWithParams("/api/destinations", {tripKey}), {method: 'POST', body:formData}).then((response)=>
         response.json()).then(locationData => {
-        updateLocations(locationData);
         updateStartDestination(locationData);
-      });
-    });
+      }); 
+    }
+    else{
+      alert("Location Invalid");
```
j-vent, line 260: It's good to also return the status code. Update this in all places where request is sent.

byway/src/main/webapp/destinations.js
```diff
@@ -134,16 +134,16 @@ function addMarker(searchBox){
 function updateLocations(locationData){
     const container = document.getElementById('destinations-container');
     container.innerText = "";
-    let destinationArray= locationData.destinations;
+    let destinationArray = locationData.destinations;
     destinationArray.forEach((destination) => {
       const request = {
-        query: destination,
+        placeId: destination,
         fields: ["name", "photos", "formatted_address", "rating", "business_status"]
```
j-vent, line 141: Seems like some of these fields are unused.

byway/src/main/webapp/destinations.js
```diff
@@ -155,18 +155,18 @@ function updateLocations(locationData){
 /** 
 * adds Users Input Destinations to DOM with image and address
 */
-function addLocationToDom(results,container){
+function addLocationToDom(place,container){
```
j-vent, line 158: It seems like container is not used in the calling function, so instead of passing it as a param just declare it inside this function directly.

renau06: I do use it in line 150 because I clear the container before adding all of the locations the user has input

byway/src/main/webapp/destinations.js
renau06: took them out!

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
renau06: fixed it

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
```diff
@@ -33,8 +33,9 @@ function initializeDestinationsPage(){
   }); 
 }
 
-const urlParams = new URLSearchParams(window.location.search);
-const tripKey = urlParams.get('tripKey');
+const tripKey = getTripKeyFromUrl();
+
+let service;
```
landondare, line 38: Nit: give this a slightly more descriptive name like `PlacesService`

zstewar1 (dismissed):

byway/src/main/webapp/destinations.js
```diff
@@ -218,21 +237,50 @@ function getCurrentAddress(){
...
+    sendPlaceId("start-search-box");
+  });
 }
 
+/** 
+* gets User Input from search box and sends the Place Id of user input to the server
+* elementName indicates which search box to get from, param indicates which param to save as
+* @param {String} elementName
+*/
+function sendPlaceId(elementName){
```
landondare, line 257: Nit: rename this to something like `savePlaceId`. You are doing more than just sending the info to the server.

byway/src/main/webapp/destinations.js
```diff
@@ -218,21 +237,50 @@ function getCurrentAddress(){
...
     event.preventDefault();
-    const formData = new  FormData(document.getElementById("user-input-form"));
-    fetch(buildUrlWithParams("/api/destinations", {tripKey}), {method: 'POST', body:formData}).then((response)=>
-        response.json()).then(locationData => {
-        updateLocations(locationData);
-        updateStartDestination(locationData);
-      });
-    });
+    sendPlaceId("destinations-search-box");
+    sendPlaceId("start-search-box");
```
landondare, line 248: Can you combine these two calls so you don't have to make 2 separate requests to the server? Also would prevent you from updating the UI twice

byway/src/main/webapp/destinations.js
```diff
@@ -218,21 +237,50 @@ function getCurrentAddress(){
...
+  };
+  service.findPlaceFromQuery(request, (results, status) => {
+    console.log(status);
+    if (status === google.maps.places.PlacesServiceStatus.OK) {
+      formData.append(elementName, results[0].place_id);
+      fetch(configureTripKeyForPath(tripKey, '/api/destinations'), {method: 'POST', body:formData})
+      .then((response)=>
+        response.json())
+      .then(locationData => {
+        if (elementName == "start-search-box"){
```
landondare, line 271: Seems like this is separate logic from the actual call to the service. It would be better to move this outside this function and separate it from the code that makes requests to the server. This could be moved to a separate function, or you could otherwise return a promise and handle the updating the UI in this functions caller.

byway/src/main/webapp/destinations.js
renau06: I've tried to combine them, but since it would have to call findPlacefromQuery two seperate times, I wasn't able to get it to work to update both the start and destination params at the same time

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
renau06: done

byway/src/main/webapp/destinations.js
renau06: moved to seperate function

byway/src/main/webapp/destinations.js
renau06: I just updated it so that savePlaceId does both calls togeth

byway/src/main/webapp/destinations.js
renau06: this function actually ended up being unnecessary so i removed it

byway/src/main/webapp/destinations.js
```diff
@@ -218,21 +237,69 @@ function getCurrentAddress(){
...
+        resolve(results);
+      }
+      else{
+        alert("Status: " + status);
+        reject(new Error("Could not find place:" + status));
+      }
+    });
+  }).catch(error => {
+    alert("Error: cannot process this request due to " + error);
+  });
```
zstewar1, line 282: When structuring code like this, where you bridge a callback-based API into a `Promise`, it's better to keep the initial promise as narrowly-scoped as possible. Don't do anything in the initial `new Promise(...)` except call the callback-based API and resolve/reject with the result, then handle any extra/external actions by chaining off of that promise.

In fact, when you have a common API which you need to bridge into a promise, I would recommend making a separate function which returns a promise-based version of the same work as the callback-based code, something like:

```js
function findPlaceFromQuery(request) {
  return new Promise(resolve => {
    placesService.findPlaceFromQuery(request, (result, status) => resolve({result, status}));
  });
}
```

That way you have a simple function which you can use to make an async query, like this:

```js
findPlaceFromQuery(destRequest)
  .then(({result, status}) => {
    // ...
  });
```

(Note that in this code, I've twice used `{result, status}`. In `resolve({result, status})` I am packing both the result and status into an object together, since `resolve` can only take one value and I want to pass both, the only way to do that is to put both `result` and `status` in an object together. In the second example block where I write `({result, status}) => { ... }`, I'm doing the reverse: I'm *unpacking* `result` and `status` from that object to get them both back as separate variables in the "then" handler.)

zstewar1, line 282: There's no point in rejecting if you're immediately going to catch your own error. Consider what this would look like if it was not using promises:

```
try {
  if (status == "OK") {
    return result; // resolve = return
  } else {
    alert("Status: " + status);
    throw new Error("..."); // reject = throw
  }
} catch (error) {
    alert(error);
}
```

Look at what happens in the error case:

1) `alert("Status: " + status);`
2) `alert(new Error("..."));` throw wrapped directly in catch + alert is basically equivalent to just sticking the error into the alert.
3) `return undefined;` Since there's no explicit return in the `catch` case, it is the same as returning `undefined`.

renau06: ok added this function

byway/src/main/webapp/destinations.js
```diff
@@ -218,21 +237,69 @@ function getCurrentAddress(){
...
+
+  promise.then(() => { 
+    placesService.findPlaceFromQuery(startRequest, (results, status) => {
```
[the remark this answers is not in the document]
renau06: fixed

byway/src/main/webapp/destinations.js
renau06: removed the reject

PR #31, opened 2020-08-19
Add utils.js

j-vent:
Utility file of shared javascript functions to be used across js files. I will use these functions in the routepage  in a separate PR.
Please ignore the linter error for JS, I'm fairly sure it's because nothing is using these functions.


byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Uses placeId to retrieve details like coordinates, place name, etc.
+ * @param {String} placeId a textual identifier that uniquely identifies a place
+ * @return {Promise} result a Place object
+ */
+function findPlace(placeId) {
+  const request = {
+    placeId: placeId,
+    fields: ['name', 'geometry', 'place_id']
```
renau06, line 23: add photos and formatted_address field so multiple pages can use

j-vent: done!

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,55 @@
...
+  const result = new Promise((resolve, reject) => {
+    placesService.getDetails(request, (result, status) => {
+      if(status == "OK") {
+        resolve(result);
+      } else {
+        alert("Status: " + status);
+        reject(result);
+      }
+    })
+  });
```
Leo2452, line 34: Remember that if it rejects the result, the promise is rejected and it needs to handle the value it rejects. After the Promise type, chain a `.catch()` promise to handle the rejected result.

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,55 @@
...
+    // arbitrary center as it will get recentered to the route 
+    center: new google.maps.LatLng(0,0)
+  }
+  map = new google.maps.Map(document.getElementById('map'), mapOptions);
+  directionsRenderer.setMap(map);
+  placesService = new google.maps.places.PlacesService(map);
+  
+}
+
+/* export findPlace, initMap */
```
Leo2452, line 54: Use exported instead of export for the eslint check

j-vent: I don't see the keyword "exported" in any javascript documentation.

byway/src/main/webapp/utils.js
j-vent: Ah right, thanks for reminding me.

byway/src/main/webapp/utils.js
Leo2452: From the ESLint docs: [](https://eslint.org/docs/rules/no-unused-vars#exported)

j-vent: Oh it's an eslint specific thing, not js. Will change.

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,42 @@
...
+    placeId: placeId,
+    fields: ['name', 'geometry', 'place_id', 'photos', 'formatted_address']
+  }
+  const result = new Promise((resolve, reject) => {
+    placesService.getDetails(request, (result, status) => {
+      if(status == "OK") {
+        resolve(result);
+      } else {
+        alert("Status: " + status);
+        reject(result);
```
zstewar1, line 31: I think argument to `reject` should be an error.

j-vent: Changed.

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,33 @@
...
+    })
+  }).catch(error => {
+      alert(error);
+  });
+
+  return result;
+}
+
+/* exported findPlace, map, placesService */
+/* global placesService:true */
```
zstewar1, line 33: I don't think you need the `:true` here. Where did you find this `name:true` syntax?

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,33 @@
...
+      if(status == "OK") {
+        resolve(result);
+      } else {
+        alert("Status: " + status);
+        reject(new Error("Could not retrieve place result object from request."));
+      }
+    })
+  }).catch(error => {
+      alert(error);
+  });
```
zstewar1, line 27: There's no point in rejecting if you're immediately going to catch your own error. Consider what this would look like if it was not using promises:

```
try {
  if (status == "OK") {
    return result; // resolve = return
  } else {
    alert("Status: " + status);
    throw new Error("..."); // reject = throw
  }
} catch (error) {
    alert(error);
}
```

Look at what happens in the error case:

1) `alert("Status: " + status);`
2) `alert(new Error("..."));` throw wrapped directly in catch + alert is basically equivalent to just sticking the error into the alert.
3) `return undefined;` Since there's no explicit return in the `catch` case, it is the same as returning `undefined`.

byway/src/main/webapp/utils.js
j-vent: I said it in the other PR, I needed to add this because I got an error saying global vars can't be modified.

j-vent:
I noticed I missed documenting a param (placesService), but I updated my code in the other PR instead since I don't want to ask Zach for a reapproval of this branch again.

PR #32, opened 2020-08-19
Route using IDs

j-vent:
This is a rewrite of routepage.js which uses Place objects/IDs instead of strings. The utils.js  file from another PR is also used, so I'm sure there will be a linter error as some variables are declared there.


byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,54 @@
+/** Script that contains shared functions, primarily for use in the third page */
```
renau06, line 1: update utils.js once its merged

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,54 @@
...
+    placeId: placeId,
+    fields: ['name', 'geometry', 'place_id']
+  }
+  const result = new Promise((resolve, reject) => {
+    placesService.getDetails(request, (result, status) => {
+      if(status == "OK") {
+        resolve(result);
+      } else {
+        alert("Status: " + status);
+        reject(result);
```
zstewar1, line 31: I think value to `reject` should be an error type.

byway/src/main/webapp/routepage.js
```diff
@@ -161,17 +154,36 @@ function getRouteOnload(){
...
-      destinations = trip.destinations;
-      route.push(...trip.route)
+      try{
+        let res = await findPlace(trip.start);
+        start = end = res.name;
+      } catch (error) {
+        console.error("Could not retrieve a start nor end point due to: ", error);
+      }
+      
+      for(let destination of trip.destinations){
```
Leo2452, line 166: I think being a bit more descriptive with the variable names would help here. Since `trip.destinations` stores the place ids, consider renaming: `destination` -> `destinationId`. This goes for the `trip.route` list as well

byway/src/main/webapp/routepage.js
```diff
@@ -229,10 +240,16 @@ function getRecsOnload() {
   clearRecs();
   fetch('/api/recs')
   .then(response => response.json())
-  .then((recommendations) => {
-     recommendations.forEach((rec)=>{
-      recs.add(rec);
-    })
+  .then(async (recommendations) => {
+    for (let rec of recommendations){
```
Leo2452, line 244: Same here, consider naming the rec variable to be id-specific

byway/src/main/webapp/routepage.js
```diff
@@ -67,7 +60,7 @@ function calcRoute() {
     origin:  start,
     destination: end,
     travelMode: 'DRIVING',
-    waypoints:  route.map(waypoint => ({location: waypoint})),
+    waypoints:  route.map(waypoint => ({location: waypoint.name})),
```
zstewar1, line 63: Can the driving directions system take the placeid as the destination ID?

byway/src/main/webapp/utils.js
j-vent: Okay, will change in the utils.js PR!

byway/src/main/webapp/routepage.js
j-vent: Yeah you're right, thanks for pointing it out. Changed.

byway/src/main/webapp/routepage.js
j-vent: Changed!

byway/src/main/webapp/utils.js
j-vent: Changed.

byway/src/main/webapp/routepage.js
j-vent: It cannot take placeId, only lat/lng, String or place object. I misspoke about using a Place object, it's really a PlaceResult so I will be using the lat/lng field instead of the String.

byway/src/main/webapp/routepage.js
```diff
@@ -247,21 +273,21 @@ function renderRecsList(){
...
     }
   });
   return recBtn;
 }
 
-/* exported initMap, generateRoute */
-/* global google */
\ No newline at end of file
+/* exported initMap, generateRoute, placesServices */
+/* global google:true, map:true, findPlace:true*/
```
zstewar1, line 293: Just listing the globals should be sufficient, where did you get `name:true` from?

j-vent: If I don't put `:true` eslint gives an error saying that global vars should not be modified.

byway/src/main/webapp/routepage.js
```diff
@@ -53,21 +57,22 @@ function initMap() {
...
     center: new google.maps.LatLng(0,0)
-    // center: new google.maps.LatLng(40.730610, -73.935242) // coordinates of NYC
   }
   map = new google.maps.Map(document.getElementById('map'), mapOptions);
   directionsRenderer.setMap(map);
+  placesService = new google.maps.places.PlacesService(map);
   
 }
-
+//  new google.maps.LatLng(0,0)
```
landondare, line 68: Nit: remove this comment?

byway/src/main/webapp/routepage.js
```diff
@@ -159,19 +164,39 @@ function clearRoute(){
 /** Get trip info from datastore onload */
 function getRouteOnload(){
   clearRoute();
-  fetch('/api/stop')
+  fetch(configureTripKeyForPath(tripKey, "/api/stop"))
   .then(response => response.json())
-  .then((trip) => {
+  .then(async (trip) => {
```
landondare, line 169: Good use of some the async functionality we discussed this week!

byway/src/main/webapp/routepage.js
```diff
@@ -159,19 +164,39 @@ function clearRoute(){
...
+        } catch (error) {
+          console.error("Could not retrieve route due to: ", error);
+        }
+      }
       calcRoute();
     }
     else{
-      console.log("Could not retrieve any routes nor destinations associated with this trip." +
-       "Please reload page and try again.");
+      console.log("Could not retrieve any routes nor destinations associated with this trip. Please reload page and try again.");
```
landondare, line 199: In a future pull request, something worth considering is converting all these `console.log` and `alert()` messages to error messages show to the user in the HTML. It would be nice to have better error handling from a user perspective.

byway/src/main/webapp/routepage.js
```diff
@@ -186,24 +211,23 @@ function renderRouteList(){
...
+  if(destinations.some(destination => destination.name === waypoint.name)){
     routeBtn.className =  "btn destination-btn";
+  } else {
+    routeBtn.className =  "btn stop-btn";
   }
   routeBtn.addEventListener("click", function() {
     // only delete if the waypoint is only a stop, not a destination
-    if(!destinations.includes(waypoint)){
-      route = route.filter(stop => stop != waypoint);
+    if(!destinations.some(destination => destination.name === waypoint.name)){
```
landondare, line 227: Can you get rid of this check and only add the event listener in the case where it is a stop? Its kind of redundant to have this check twice, and the event listener doesn't do anything for the destination buttons anyways.

byway/src/main/webapp/utils.js
```diff
@@ -0,0 +1,32 @@
...
+ * @return {Promise} result a Place Result object with fields name, geometry, id, etc.
+ */
+function findPlace(placeId) {
+  console.log(placeId);
+  const request = {
+    placeId: placeId,
+    fields: ['name', 'geometry', 'place_id', 'photos', 'formatted_address']
+  }
+  const result = new Promise((resolve, reject) => {
+    placesService.getDetails(request, (result, status) => {
```
landondare, line 18: Since placesService is not defined in this file, it probably makes better sense to pass it as a parameter rather than rely on it being a global variable.

byway/src/main/webapp/routepage.js
j-vent: Oops, didn't see that. Thanks!

byway/src/main/webapp/routepage.js
j-vent: Thanks! It helped out because I wasn't sure if this was proper.

byway/src/main/webapp/routepage.js
j-vent: Do you mean have a field to display on the HTML? Why is bad to have these alert messages instead? Will keep this in mind, thanks.

byway/src/main/webapp/routepage.js
j-vent: Thanks for catching that, it was redundant. Changed!

byway/src/main/webapp/utils.js
j-vent: I changed it, but what's wrong with it being a global variable? Is it more error prone this way?

byway/src/main/webapp/routepage.js
zstewar1: `/* global varName */` defaults to treating values like `const`, but can be changed to treat them like `let` with `/* global varName:writeable */`. If you need to make a global writeable, prefer to use `:writeable` rather than `:true` because that is easier to understand.

However: if you are going to be initializing some global variable in a particular file, then you should declare the variable in the same file with `let`, and "import" it using `/* global */` in other files where you only need to use the value, rather than need to assign it. If a particular file handles setting up the value for a global, then it makes sense to also *declare* the global in that file (and mark it as exported). Having declaration and initialization separate is error-prone, since the declaration and assignment are separated its easier to mess up the naming. Also, if you get in the habit of using `:writeable` globals, you may accidentally overwrite globals that you didn't intend to; if you only assign them in the same file where they are declared, then in all other files, the linter will help you make sure that you are not overwriting them by accident.

byway/src/main/java/com/google/sps/servlets/StopsServlet.java
```diff
@@ -41,14 +40,17 @@
...
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Trip trip;
+
     try {
-      trip = Trip.fromEntity(datastore.get(key));
-    } catch (EntityNotFoundException e) {
+      // trip = Trip.fromEntity(datastore.get(key));
+      trip = Trip.getTrip(datastore, (String) request.getParameter("tripKey"));
+
+    } catch (NullPointerException e) {
```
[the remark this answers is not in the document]
j-vent: Changed to just check for null, as per your suggestion.

PR #34, opened 2020-08-20
Route styling

j-vent:
Styling for the route page, added email and finish buttons.


byway/src/main/webapp/routepage.css
```diff
@@ -1,104 +1,144 @@
+@import url('https://fonts.googleapis.com/css2?family=Noto+Sans+JP:wght@700&display=swap');
+ 
 #map{
-  height:500px; /* change to 40 vw */
-  width:500px; /* 5 vw padding*/
+  height: 750px; 
+  width: 700px;
```
Leo2452, line 5: The map container should align with Rena's from destinations.css, right? You could also call the css file inside routepage.html to not rewrite.

j-vent: I changed the sizing for aesthetic appeal, since we don't share the map object amongst these pages I'm just going to leave it.

PR #35, opened 2020-08-20
Add home page

renau06:
Adds home page that displays user past trips adn button to create new, adds logout link in header, uses findPlace from util.js

This PR incorporates the changes from util.js and my other PR Change to Place IDs, but I am sending it now because I want to have enough time to finish it. I will merge the changes into this once the other PRs are sent.


byway/src/main/webapp/home.js
```diff
@@ -1,3 +1,13 @@
+/* global google, configureTripKeyForPath, setLogoutLink*/
+/* exported placesService*/
+
+let placesService;
+
+if(!!window.performance && window.performance.navigation.type == 2)
+{
+    window.location.reload();
+}
```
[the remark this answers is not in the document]
renau06: If the user navigates to the page via the back button, this reloads the page. I did this because if user created a new trip and then clicked the back button to go back to home page, I wanted that new trip to appear on their list of trips

byway/src/main/java/com/google/sps/data/Trip.java
```diff
@@ -133,7 +133,7 @@ public void addDestination(String destination) {
     destinations.add(destination);
   }
 
-  /* Adds a destination point of the trip as plain text. */
+  /* Adds a route point of the trip as plain text. */
```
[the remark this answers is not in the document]
renau06: done

byway/src/main/webapp/destinations.js
```diff
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 /* exported initAutocomplete, getCurrentAddress */
-/* global google, setProgressBar, getTripKeyFromUrl, configureTripKeyForPath */
+/* global google, setProgressBar, getTripKeyFromUrl, configureTripKeyForPath, setLogoutLink, findPlace, map:true */
```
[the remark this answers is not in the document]
renau06: fixed, since justine removed the delcaration of map from util.js, I made map a global variable only in destinations.js

byway/src/main/webapp/home.js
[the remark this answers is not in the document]
zstewar1: Please use the named constant `PerformanceNavigation.TYPE_BACK_FORWARD` rather than a numeric constant/magic number.


Did you run into problems with this, it seems very odd to me that hitting 'back' would ot result in loading content on the previous page again.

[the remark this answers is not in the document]
renau06: Yeah i did have problems. Without this, when I hit the back button, it wouldn't add the new trip until I refreshed the page.

byway/src/main/webapp/base.js
```diff
@@ -15,6 +16,23 @@ function setProgressBar(pageNumber){
   document.getElementById('view-route-progress-bar').href = configureTripKeyForPath(tripKey, '/routepage.html');
 }
 
+function setLogoutLink(){
```
zstewar1, line 19: Since this also redirect when the user is *not* logged in, you may want to choose a different name. Keep in mind that `set` usually means 'assign to a field', where this is more like 'setup logout link'; set = simple assignment, setup = may do some other work. Also, since this does a redirect, it should have a doc comment that mentions that.

renau06: fixed

byway/src/main/webapp/home.js
```diff
@@ -6,11 +17,107 @@ else{
...
   document.getElementById('create-trip').addEventListener('click', () => {
     fetch('/api/createtrip', {method: 'POST'}).then((response) => response.json()).then((trip) =>{
-      window.location.href = '/destinations.html?tripKey=' + trip.keyString;  
+      let tripKey = trip.keyString;
+      window.location.href = configureTripKeyForPath(tripKey, "/destinations.html");
     });
   });
 }
 
+function createPastTrip(){
```
landondare, line 30: Nit: it may be more accurate to call this `loadPastTrips` or something along those lines. "createPastTrip" is a bit confusing, it sounds like you are creating new trips from past info something. This doesn't really create trips, so the "create" part is not strictly accurate.

byway/src/main/webapp/home.js
```diff
@@ -6,11 +17,107 @@ else{
...
+  fetch('/api/gettrips').then((response) => response.json()).then((tripIds) => {
+    let tripNum = 1;
+    tripIds.forEach(trip => {
+      let isDestinationsMissing = trip.destinations.length == 0;
+      let isInterestsMissing = trip.interests.length == 0;
+      if(isDestinationsMissing|| isInterestsMissing) {
```
landondare, line 36: Seems like you could just do this check inline rather than use two variables. In general, if you have a boolean value that is only read once, you can just get its value where it is used.

byway/src/main/webapp/home.js
renau06: changed to load

byway/src/main/webapp/home.js
renau06: I passed these variables into the showIncompleteTrip so i kept them as variables, but I changed to an inline check

landondare: If they are used elsewhere you can ignore my comment. Either way is fine.

PR #36, opened 2020-08-20
Customize potential interests

Leo2452:
Make an explicit list of noninterests to filter from the PlaceTypes given through the maps Places API. Additionally, make another list of interests to add on top of the acceptable interests.

If anyone else thinks of other interests, please comment in the PR

### DIFF
--- a/byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
+++ b/byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
@@ -17,6 +17,7 @@ package com.google.maps;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceConfig;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.common.collect.ImmutableSet;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import com.google.maps.model.PlaceType;
@@ -24,9 +25,7 @@ import com.google.sps.data.Trip;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -51,23 +85,29 @@ public void init() {
...
+    ArrayList<String> places = new ArrayList<String>();
     for (PlaceType location : PlaceType.values()) {
-      String place = formatLocation(location);
-      places.add(place);
+      if (!nonInterests.contains(location.toString())) {
+        String place = formatPlace(location.toString());
+        places.add(place);
+      }
     }
+    for (String customInterest : customInterests) {
```
j-vent, line 95: You can use `list1.addAll(list2)` instead of having to iterate through the entire arraylist.

Leo2452: At first, I didn't do this since I wanted to keep `customInterests` in the same format as the other lists (i.e. underscores as spaces, non-capitalized, etc.), and then format them later in similar fashion. I guess since they are being hard-coded anyways, I can format it and use this method to avoid formatting through a function call.

Fixed

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -51,23 +90,26 @@ public void init() {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    ArrayList<String> places = new ArrayList<>();
+    ArrayList<String> places = new ArrayList<String>();
     for (PlaceType location : PlaceType.values()) {
-      String place = formatLocation(location);
-      places.add(place);
+      if (!NON_INTERESTS.contains(location.toString())) {
```
zstewar1, line 95: if NON_INTERESTS is always going to be matched against `PlaceType` values, consider using ImmutableSet<PlaceType> rather than ImmutableSet<String>. That way there's no mistaking what format the contents are, and you don't have to worry about values names changing. 

Note that if you change to `PlaceType`, you should use `Sets.immutableEnumSet` instead of `ImmutableSet.of`.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
```diff
@@ -39,6 +41,43 @@
   /** {@link Type} of an {@link ArrayList} containing {@link String}, for gson decoding. */
   private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
 
+  /**
+   * Formatted to be all lowercase, delimited by an underscore. Follows the format used for
+   * PlaceType objects.
+   */
+  private static final ImmutableSet<String> NON_INTERESTS =
```
landondare, line 48: Can you name this something slightly more descriptive? hard to know what a non-interest is in this context. Something like NON_INTEREST_PLACETYPES or something that otherwise describes what these values are in the larger context.

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java

byway/src/main/java/com/google/sps/servlets/PlacesServlet.java
Leo2452: Updated both name and its description

PR #37, opened 2020-08-21
Check valid trip key

Leo2452:
Redirect the user to the homepage if trip key is not found immediately when searching the url, rather than checking when attempting to use the trip key.


byway/src/main/webapp/base.js
```diff
@@ -12,29 +12,29 @@ function setProgressBar(pageNumber){
...
+ * and have a valid trip key to reference.
  * @return String of trip key
  */
 function getTripKeyFromUrl() {
-  return new URLSearchParams(location.search).get('tripKey');
+  const tripKey = new URLSearchParams(location.search).get('tripKey');
+  if(tripKey === null) {
+    alert("tripKey not created! Create a new trip.");
+    window.location.href = '/index.html';
+    return "";
```
landondare, line 25: Probably makes more sense to return undefined here if anything.

byway/src/main/webapp/base.js
```diff
@@ -12,29 +12,29 @@ function setProgressBar(pageNumber){
...
+ * Go through url to retrieve the trip key. If the trip key is null,
+ * redirects the user to the page where they can make a new trip
+ * and have a valid trip key to reference.
  * @return String of trip key
  */
 function getTripKeyFromUrl() {
-  return new URLSearchParams(location.search).get('tripKey');
+  const tripKey = new URLSearchParams(location.search).get('tripKey');
+  if(tripKey === null) {
+    alert("tripKey not created! Create a new trip.");
```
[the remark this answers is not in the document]
Leo2452: Fixed

PR #39, opened 2020-08-24
Add email function

j-vent:
Added a servlet for the emails as well as some js functions to send out a route email.


byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,34 @@ function createRecButton(rec){
...
+   return routeLink;
+}
+
+
+function sendEmail(){
+  fetch('/api/email')
+  .then(response => response.json())
+  .then(email =>{ 
+    let emailLink = "mailto:" + email + "?subject=Your%20Roadtrip%20Plan&body=Thank%20you%20for%20using%20Byway!%20" +
+      "Your%20destinations%20are%20listed%20below.%20Click%20the%20link%20to%20see%20your%20route%20in%20Google%20Maps.%0D%0A"
```
renau06, line 315: small nit: this says you destinations are listed below, so I think we should either list the destinations as well or take out that sentence

j-vent: I just said "route" to be more general, thanks for pointing it out!

byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,34 @@ function createRecButton(rec){
...
+    routeLink += "&origin=" + start.name.split(' ').join('%20') + "&destination=" + end.name.split(' ').join('%20') +"&waypoints=";
+    for(let place of route){
+      console.log(place);
+      routeLink += place.name.split(' ').join('%20') + "|";
+    }
+   return routeLink;
+}
+
+
+function sendEmail(){
```
Leo2452, line 309: Add a small description of what this function does/includes

j-vent: Thanks, added a comment!

byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,35 @@ function createRecButton(rec){
...
+      routeLink += place.name.split(' ').join('%20') + "|";
+    }
+   return routeLink;
+}
+
+/** Opens a new window of an email client with prefilled to, subject, and body field 
+ *  containing a Google Maps link of the user's route 
+ */
+function sendEmail(){
+  fetch('/api/email')
```
landondare, line 313: Why not just let the user fill out an email address themselves? What if they want to mail it to someone else? Seems easier to have a box where they can enter an email address and then generate the email rather than getting it from the user data on the server.

byway/src/main/java/com/google/sps/servlets/EmailServlet.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that handles data for emails */
+@WebServlet("/api/email")
+public final class EmailServlet extends HttpServlet {
```
landondare, line 29: This seems a bit overkill to get the users email address.

byway/src/main/webapp/routepage.js
j-vent: I added a field instead.

byway/src/main/java/com/google/sps/servlets/EmailServlet.java
j-vent: Yeah, I think pre-filling the email field could work but I know what you mean. I don't use it but haven't yet deleted this servlet.

byway/src/main/java/com/google/sps/servlets/EmailServlet.java
```diff
@@ -0,0 +1,47 @@
...
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+  private final Gson gson = new Gson();
+  private final UserService userService = UserServiceFactory.getUserService();
+
+  /* Passes saved trip object to show route in the schedule panel, start/end in request */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userEmail = userService.getCurrentUser().getEmail();
```
zstewar1, line 38: What about if the user isn't logged in (CurrentUser == null)?

byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,45 @@ function createRecButton(rec){
...
+
+/** Creates a URL link to Google Maps based on the start/end point and route
+ *  @returns {String} routeLink url containing query params for the user’s route
+ */
+function generateRouteLink(){
+    let routeLink = "https://www.google.com/maps/dir/?api=1&travelmode=driving" 
+    routeLink += "&origin=" + start.name.split(' ').join('%20') + "&destination=" + 
+      end.name.split(' ').join('%20') + "&waypoints=";
+    for(let place of route){
+      routeLink += place.name.split(' ').join('%20') + "|";
```
zstewar1, line 304: What about other characters that need to be escaped besides space? There should be a function that does this encoding for you (`encodeURIComponent`). Also, you should never build a complexly formatted string by manually concatenating values together if there is a tool that can build it for you: 

```
new URLSearchParams({
    api: 1,
    travelmode: 'driving',
    destination: 'wherever the place',
    origin: 'from somewhere?',
    waypoints: ['a place', 'b somewhere ??? along the way? @%'].join('|')
}).toString()

// outputs: "api=1&travelmode=driving&destination=wherever+the+place&origin=from+somewhere%3F&waypoints=a+place%7Cb+somewhere+%3F%3F%3F+along+the+way%3F+%40%25"
```

Anytime you need to build search params for a URL, just use UrlSearchParams with an object arg to specify the parameters, so you don't have to worry about escaping them or messing up the encoding.

byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,45 @@ function createRecButton(rec){
...
+    }
+   return routeLink;
+}
+
+/** Determines if email address is valid using regex patterns
+ *  @param {String} email email address 
+ *  @returns {boolean} is the email a valid email address
+ */
+function validateEmail(email) {
+    const re = /^\w+([-]?\w+)*@\w+([-]?\w+)*(\.\w{2,3})+$/;
```
zstewar1, line 314: Email is regex is overly strict, bans longer domain names like `.info`, and doesn't support emails with `+` and other chars in the name.

You could look at https://www.regular-expressions.info/email.html for a regex that matches email addresses more generally.

byway/src/main/webapp/routepage.js
```diff
@@ -292,5 +292,45 @@ function createRecButton(rec){
...
+    return re.test(String(email).toLowerCase());
+}
+
+/** Opens a new window of an email client with prefilled to, subject, and body field 
+ *  containing a Google Maps link of the user's route 
+ */
+function sendEmail(){
+  let emailAddress = document.getElementById("email-address").value;
+  let emailLink = "mailto:" + emailAddress + "?subject=Your%20Roadtrip%20Plan&body=Thank%20you%20for%20using%20Byway!%20" +
```
zstewar1, line 323: Please don't write manually encoded values like this, too easy to mess up.

Also, since these are query parameters again, you can use `URLSearchParams` here too.

byway/src/main/java/com/google/sps/servlets/EmailServlet.java
j-vent: I might just remove this servlet since it's unused.

byway/src/main/webapp/routepage.js
j-vent: Alright, I used a regex pattern listed on that site which supports 99% of email addresses.

byway/src/main/webapp/routepage.js
j-vent: Changed, I took out encodeURIComponent() since you said urlSearchParams() handles encoding.

byway/src/main/webapp/routepage.js
j-vent: Changed to use URLSeachParams()

PR #41, opened 2020-08-25
Fixes to Home Page 

renau06:
add delay to fix Over query limit and add comments to functions


byway/src/main/webapp/home.js
```diff
@@ -110,7 +148,14 @@ function calcRoute(directionsService, directionsRenderer, start, end, waypoints)
   directionsService.route(request, function(response, status) {
     if (status == 'OK') {
       directionsRenderer.setDirections(response);
-    } else {
+    } 
+    else if (status === google.maps.DirectionsStatus.OVER_QUERY_LIMIT) {
+      delay++;
+      setTimeout(function () {
+        calcRoute(directionsService,directionsRenderer, start, end, waypoints);
+      }, delay * 1000);
```
Leo2452, line 156: Might want to use delayPromise function from recommendations.js, you can move into base.js and align/sync in html/js files as needed for eslint errors

renau06: okay changed to use delayPromises. For now I am keeping the delayPromises function in both home.js and recomendations.js until we decide what to do with base.js and utils.js

byway/src/main/webapp/home.js
```diff
@@ -27,6 +31,9 @@ function initializeHomePage(){
   });
 }
 
+/**
+ * For each of user's past trips, creates a div with either a complete trip or a trip missing some inputs
+ */
 function loadPastTrip(){
   fetch('/api/gettrips').then((response) => response.json()).then((tripIds) => {
```
j-vent, line 38: Separate .then into their own line.

renau06: done!

byway/src/main/webapp/home.js
```diff
@@ -110,9 +152,16 @@ function calcRoute(directionsService, directionsRenderer, start, end, waypoints)
   directionsService.route(request, function(response, status) {
     if (status == 'OK') {
       directionsRenderer.setDirections(response);
-    } else {
+    } 
+    else if (status === google.maps.DirectionsStatus.OVER_QUERY_LIMIT) {
+      delayPromise(1000)
+      .then(() => calcRoute(directionsService,directionsRenderer,start,end,waypoints));
```
zstewar1, line 158: Writing this as a recursive callback loop makes the logic harder to read and understand, and potentially harder to control, e.g. it's difficult to set a limit to the number of attempts that will be made. It's also impossible for the caller to tell if the function has completed or not, since calling `calcRoute` doesn't wait for the route to be calculated in any way.

I would recommend making a function that returns a promise of the result of `directionService.route`, and write `calcRoute` as an async function with a loop that retries the call to the directions service until it works (or some predetermined attempt limit is reached).

renau06: done

byway/src/main/webapp/home.js
```diff
@@ -96,23 +126,76 @@ function initMap(start, end, route, keyString) {
...
+    try{
+      let result = await getDirections(directionsService, start, end, waypoints)
+      directionsRenderer.setDirections(result);
+      return;
+    }catch(error){
+      if (error.status === google.maps.DirectionsStatus.OVER_QUERY_LIMIT){
+        await delayPromise(1000);
+      }
+    }
+  }
```
zstewar1, line 151: If you run out of tries, it would be good to at least log an error, if not throw an exception.

PR #43, opened 2020-08-25
Routepage UI

Leo2452:
Includes alerts to user through html (and resetting), removing markers, and a loading spinner while recommendations are loading


byway/src/main/webapp/routepage.html
```diff
@@ -6,16 +6,24 @@
     <link rel="stylesheet" href="/bootstrap/bootstrap.min.css">
```
j-vent, line 6: Choose to use bootstrap as either CDN or move the dowloaded version (preferred by Landon) into this directory, not both.

Leo2452: Fixed, used the downloaded version

zstewar1 (dismissed):

byway/src/main/webapp/recommendations.js
```diff
@@ -164,14 +165,32 @@ function findPlacesWithTextSearch(request, statuses) {
...
+  const allStatuses = formatStatusMessages(statuses);
+  msgContainer.innerText = "Showing limited results due to: " + allStatuses;
+}
+
+/**
+ * Go through status messages and add punctuation to make a sentence.
+ * @param {Set} statuses contains elements with status codes from placesService
+ * @return String with punctuation as it lists all status codes.
+ */
+function formatStatusMessages(statuses) {
```
landondare, line 179: Trying to add proper punctation to the individual status messages is not necessary and adds clutter to your code. You could do something much simpler like delineate them with new lines. e.g.

```javascript
errorsStr = Array.from(statuses).join("\n")
msgContainer.innerText = "Showing limited results due to: \n" + errorsStr;
```

byway/src/main/webapp/recommendations.js
```diff
@@ -68,12 +73,7 @@ function addMainStopsToRegions() {
 /* Resets the alerts found in a previous attempt to load recommendations and reveals loading bar. */
 function resetUserAlerts() {
   document.getElementById("message-container").style.visibility = 'hidden';
-  document.getElementById("general-message").style.visibility = 'hidden';
-  let statusesContainer = document.getElementById("statuses");
-  while(statusesContainer.hasChildNodes()) {
-    statusesContainer.removeChild(statusesContainer.firstChild);
-  }
-  statusesContainer.style.visibility = 'hidden';
+  document.getElementById("loading").style.visibility = 'visible';
```
landondare, line 76: Why do you want to show the loading indicator when resetting the alerts? Is the loading indicator something you want to show by default? It would be better to show and hide the loading element only when you are actually loading something. Else your page or logic might end up in a state that could be confusing to the user.

byway/src/main/webapp/recommendations.js
```diff
@@ -224,4 +243,13 @@ function placeMarker(place) {
...
   });
+  markers.push(marker);
+}
+
+/* Iterate through all markers and set their maps to null. Make Markers array empty. */
+function clearMarkers() {
+  for (let i = 0; i < markers.length; i++) {
+    markers[i].setMap(null);
+  }
+  markers = [];
```
landondare, line 254: You are iterating through every marker and resetting the map, but then you are reassigning the value of markers? Why do you need to reset the map of the marker objects if you are just resetting the array anyways? 

If the objects inside the map are going to be discarded, then it doesn't make sense to go through and edit them first. Unless a copy of the object are held in some other data structure, in which case, there are ways you could make this more clear in the code. Without any context, this code looks like it doing something pointless.

byway/src/main/webapp/recommendations.js
Leo2452: Thanks for the suggestion, fixed

byway/src/main/webapp/recommendations.js
Leo2452: On second thought, there doesn't seem to be a reason to reset user alerts, since the only way to reload recommendations (as of now) would be to refresh the page, which gets handled with default values. I removed this for now, but will consider adding a "try again?" button as you suggested. The google maps API doesn't specify if results will always be different if the same query is passed in, but from some past experimentation the results seem to always be the same.

Updated

byway/src/main/webapp/recommendations.js
Leo2452: Setting the map to null removes the marker object from the google.maps object, but it still lived in the markers array. I still had to set it to an empty array to completely delete it (https://developers.google.com/maps/documentation/javascript/markers#remove)

I updated this function to either show or hide markers based on a boolean that is passed in. We will use this to add a toggle feature to either show markers or hide them on click events.

PR #45, opened 2020-08-26
Load preselected interests

Leo2452:
Maintain the state of interests selected in a trip. Create new servlets to handle loading interests (from Places API, in PlacesServlet) and to handle updating the interests selected for the trip (in InterestsServlet). Update status on UI as needed

### DIFF
--- a/byway/src/main/java/com/google/sps/servlets/InterestsServlet.java
+++ b/byway/src/main/java/com/google/sps/servlets/InterestsServlet.java
@@ -1,0 +1,76 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.maps;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceConfig;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import com.google.maps.model.PlaceType;
+import com.google.sps.data.Trip;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that stores which interests the user selected for their trip in datastore. Also updates the interests in POST requests. */
+@WebServlet("/api/interests")
+public final class InterestsServlet extends HttpServlet {
+
+  /** {@link Type} of an {@link ArrayList} containing {@link String}, for gson decoding. */
+  private static final Type ARRAYLIST_STRING = new TypeToken<ArrayList<String>>() {}.getType();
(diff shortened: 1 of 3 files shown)


byway/src/main/java/com/google/sps/servlets/InterestsServlet.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String tripKeyString = request.getParameter("tripKey");
+    ArrayList<String> interests = gson.fromJson(request.getReader(), ARRAYLIST_STRING);
+    Trip trip = Trip.getTrip(datastore, tripKeyString);
+    if (trip == null) {
+      response.setStatus(response.SC_NOT_FOUND);
+      return;
+    }
```
j-vent, line 71: Add `response.setStatus(HttpServletResponse.SC_NO_CONTENT);` when nothing is returned as JSON

byway/src/main/webapp/track-interests.js
```diff
@@ -58,13 +54,33 @@ function updateStatus(place, elem) {
...
+ */
+function getTripInterests() {
+  return new Promise(resolve => {
+    let tripInterests = new Set();
+    fetchInterests(tripKey)
+    .then((trip) => {
+      for (let interest of trip.interests) {
+        tripInterests.add(interest);
+      }
+      resolve(tripInterests);
```
j-vent, line 71: Might be useful to have a reject(new Error ...) which you can catch in your function await function call.

byway/src/main/java/com/google/sps/servlets/InterestsServlet.java
Leo2452: Updated

byway/src/main/webapp/track-interests.js
Leo2452: I don't think any reject is necessary here since it is just loading/saving the data from the backend. If the trip.interests is empty/null, it would resolve null and return an empty Set. I could add a check for the reject case, but that would imply there was an actual error with the Error argument, when it really just means it's a new trip with no interests yet. 

I don't think every Promise needs to handle both cases, so I think I will leave this as is since there is no real "error" to reject.

byway/src/main/java/com/google/sps/servlets/InterestsServlet.java
```diff
@@ -0,0 +1,76 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String keyAsString = request.getParameter("tripKey");
+    Trip trip = Trip.getTrip(datastore, keyAsString);
+    if (trip == null) {
+      response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+      return;
+    }
+    response.setContentType("application/json");
+    response.getWriter().println(gson.toJson(trip));
+  }
```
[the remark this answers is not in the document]
Leo2452: Fixed to be symmetrical and only get/set interests for the trip

byway/src/main/webapp/track-interests.js
```diff
@@ -58,13 +54,33 @@ function updateStatus(place, elem) {
...
 }
 
+/**
+ * Finds the list of interests that were set for the trip.
+ * If it is a new trip, no interests are listed. If the trip
+ * had previously-saved interests, return them.
+ * @return fulfilled promise containing the list of interests
+ */
+function getTripInterests() {
+  return new Promise(resolve => {
```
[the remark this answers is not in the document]

byway/src/main/webapp/track-interests.js
```diff
@@ -29,42 +30,53 @@ if(document.readyState === 'loading') {
...
+ * had previously-saved interests, return them.
+ * @return fulfilled promise containing the list of interests
+ */
+async function getTripInterests() {
+  await fetchInterests(tripKey)
+  .then((interests) => {
+    for (let interest of interests) {
+      interestsSaved.add(interest);
+    }
+  });
```
zstewar1, line 69: Try to avoid mixing async/await with promise-chaining. Both do the same thing, so just use one or the other.
Reminder these do the same thing:

```js
function getTripInterests() {
  return fetchInterests(tripKey)
    .then((interests) => {
      for (let interest of interests) {
        interestsSaved.add(interest);
      }
    });
}

async function getTripInterests() {
  let interests = await fetchInterests(tripKey);
  for (let interest of interests) {
    interestsSaved.add(interest);
  }
}
```

PR #46, opened 2020-08-27
Routepage changes

j-vent:
Made the following changes: open in Google Maps button, hide stops after selecting them on the route page, used placeIds along with placeName in the query, used placeId to identify place buttons on the lists.

Also just added the instructions to all pages here, since it's just html/css changes.


j-vent:
Will add back "deleted files" when merging with master, I just don't want to have other files reviewed.

byway/src/main/webapp/routepage.html
```diff
@@ -34,6 +34,12 @@
       <p id="general-message">Showing limited results due to: </p>
       <p id="statuses"></p>
     </div>
+
+    <div class = "tips">
```
renau06, line 38: maybe add something about how to remove stops in the tips

j-vent: Good point, added!

byway/src/main/webapp/routepage.js
```diff
@@ -317,10 +325,16 @@ function sendEmail(){
   } else{ // TO DO: Use alerts to notify user
     console.log("Please enter a valid email address.");
   }
+
+}
+
+/** Opens route on Google Maps */
+function openInGMaps(){
+  window.open(generateRouteLink());
```
zstewar1, line 333: For URLs it's preferable to update an `<a href=>` whenever the link changes rather than set an on-click handler, because onclick doesn't provide the same set of allowed operations:

onclick:
- click button to open however the default `window.open` handler opens it.

href:
- left click to open in this window.
- middle click to open in a new window/tab
- right click for menu options such as "copy link address"
- hover with mouse to see the URL that clicking will take you to.

j-vent: Changed, thanks for explaining the difference I wasn't aware of the special operations href had.

zstewar1: If you update the href in an onclick handler, then the other functionality you get from using href does apply if you haven't previously clicked the button, since the href will still be the default value. You would want to update the href whenever the origin, destination, or waypoints change.

j-vent: I know what you're saying, but I've tested it and it seems that the function is fast enough to keep up with the current route and assign it on click.

zstewar1: Did you test *all* interactions?

https://jsfiddle.net/rzu2p0va/

Look at this example, with just a link with a click handler. Click "run" before doing each of these tests (to reset the link, since once you click it, the URL will be changed):

- Hover over the link: `example.com`, but if you click it takes you to `google.com` (so the link inaccurately shows the URL it's pointed at)
- Middle click: goes to `example.com`, so the URL is *not* updated. In the context of your trip generator, that would mean the route you see if you *click* the link is different from the route you see if you *middle click* the link.
- Right click, copy URL, paste in new tab: same as middle click; it will copy an old URL to an old route from the *last* time you clicked the button.

j-vent: You're right, I only tested the two main interactions (left and right click), I changed it so the link is updated with changes to the route. Thanks!

PR #48, opened 2020-08-28
General Refactoring

j-vent:
Refactored routepage.js and recommendations.js to simplify a large, redundant request to the Directions API. Wrote helper functions to combine commonalities using promises instead of callbacks,  and to separate out different function calls. 

Also placed css and js files into subfolders for better organization, so they show up as new files. Changes were made to routepage.js and recommendations.js only.


byway/src/main/webapp/js/recommendations.js
```diff
@@ -29,32 +29,29 @@ const MAX_RECOMMENDATIONS = 1;
...
-      findRegions(result);
-      loadRecommendations();
-      orderWaypoints(result);
-      updateDistanceTime(result);
-    } else {
-      alert("Could not calculate route due to: " + status);
-    }
-  });
+  getRecommendations();
+  updatePageInfo();
```
Leo2452, line 54: updatePageInfo calls getRouteForTrip, but so does getRecommendations, so it makes two uses of the directions API. Even though this double-call seems to happen once initially, since we getRecs once, might be worth looking into.

j-vent: Yeah... I noticed this as well but I wasn't sure how to resolve it since they both need to use the directionsServices but are called at different times. I might wait on future suggestions from Zach and Landon.

j-vent:
I did not use that command, I just manually coped them and then I did `git rm` on the old files. I just checked and it looks like they're still there. I'm working on some other changes so I will fix the issue of duplicates next week.

j-vent:
Refactoring is in another PR currently. File organization will be moved into another PR, as it interferes with merging. Will use git mv in the future to prevent this.

PR #49, opened 2020-08-28
Interests searchbar

Leo2452:
Search bar to filter through the list of interest buttons. State of interests chosen is maintained (blue / grey), both during the search and when removing all letters from the search bar.


byway/src/main/webapp/track-interests.js
```diff
@@ -92,6 +93,22 @@ function createButtonForPlace(place) {
...
+/* Hide or display list of interests to match user input as a zero-indexed substring. */
+function filterInterests() {
+  const searchField = document.getElementById("search-interest");
+  const searchValue = searchField.value.toLowerCase();
+  const interestBtnsElement = document.getElementById("interests-buttons");
+  let interestBtns = interestBtnsElement.getElementsByClassName("interestBtn");
+  for (let i = 0; i < interestBtns.length; i++) {
+    const btnValue = interestBtns[i].innerText;
+    if (btnValue.toLowerCase().indexOf(searchValue) === 0) {
+      interestBtns[i].style.display = "";
```
j-vent, line 105: Seems weird to clear the display field instead of replacing it with another value to show that it's shown to the user. Try inline or inline-block.

byway/src/main/webapp/track-interests.js
```diff
@@ -92,6 +93,22 @@ function createButtonForPlace(place) {
   return button;
 }
 
+/* Hide or display list of interests to match user input as a zero-indexed substring. */
+function filterInterests() {
+  const searchField = document.getElementById("search-interest");
+  const searchValue = searchField.value.toLowerCase();
```
landondare, line 99: You may not want to filter anything out if the search value is empty.

Leo2452: It would be a nice check to be done initially, but consider this: a user types in "a," it filters for the interests starting with "a," and then the user deletes the "a" - say he meant to type in "b" instead. Since this function is called per `keyup` event, `filterInterests` would see no input and return to avoid the for loop. It would skip the filtering, which is in charge of either hiding or showing elements. It would be stuck displaying only interests starting with "a" despite there being no input, which should show all elements.

byway/src/main/webapp/track-interests.js
```diff
@@ -92,6 +93,22 @@ function createButtonForPlace(place) {
...
+  const interestBtnsElement = document.getElementById("interests-buttons");
+  let interestBtns = interestBtnsElement.getElementsByClassName("interestBtn");
+  for (let i = 0; i < interestBtns.length; i++) {
+    const btnValue = interestBtns[i].innerText;
+    if (btnValue.toLowerCase().indexOf(searchValue) === 0) {
```
zstewar1, line 104: If you check `> 0` rather than `=== 0`, you'd be able to search for a partial name, rather than only a prefix.

Leo2452: I originally had `> -1`, so that it could also account for being equal to 0, but changed it to the prefix only since typing in `a` still showed a result like `city hall` despite it not starting with`a`. You have a point about it being a partial name, i.e. searching `park` should reveal `RV park`, so I updated despite the match being anywhere in the word.

I could update it to split complete words, i.e. an interests delimited by a space such as `RV park`, and then search for a prefix per beginning of each word. But, this introduces a nested for loop when only using `> -1` accomplishes the same task, albeit not as strictly. I will leave it as is with `> -1` and if you/Landon recommend the split/for loop approach, or some other approach, I can implement.

landondare: This is an entirely subjective question about how you want your search/filter feature to work. I personally favor the partial match over the prefix approach. This seems more flexible to me. I don't think it is an issue if a search for a single letter has many results. I think it is assumed that the user will type more details in order to get more specific results. E.g. They may type "T" get too many results, and then continue typing "Thea" for Theater  etc. 

You could also split by word, which is better than searching for a strict substring. This is slightly more flexible and would be a better user experience. I wouldn't worry about the nested loop aspect. It is pretty trivial to split/ parse such simple strings. Though in this case I think it is better to stick with the more general approach of checking any partial matches over just prefixes

PR #50, opened 2020-08-28
Add titles to home page

renau06:
Add destinations as titles for Your Trips Page


byway/src/main/webapp/home.js
```diff
@@ -86,22 +87,44 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
   title.href = configureTripKeyForPath(trip.keyString, "/routepage.html");
 }
 
+async function constructTripTitle(trip, keyString){
+  let title = document.getElementById('title-' + keyString);
+  title.innerText = "";
+  for(let i = 0; i<20; i++){
+      try{
+        const promises = trip.destinations.map(async (destination) => {
+            let placeInfo = await findPlace(destination, placesService);
```
j-vent, line 113: It might be good to use try-catch blocks here and in other places using `await ` in case the promise is rejected.

byway/src/main/webapp/home.js
```diff
@@ -159,13 +182,13 @@ function delayPromise(delayMs) {
 /**
  * Class used to throw error in getDirections() with both message and status
  */
-class MapStatusError extends Error {
+/*class MapStatusError extends Error {
```
j-vent, line 185: nit: remove comment

renau06: done

byway/src/main/webapp/home.js
```diff
@@ -86,22 +87,44 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
   initMap(trip.start, trip.start, trip.route, trip.keyString);
-  title.innerText = "Trip #" + tripNum;
+  await constructTripTitle(trip, trip.keyString);
   title.href = configureTripKeyForPath(trip.keyString, "/routepage.html");
 }
 
+async function constructTripTitle(trip, keyString){
+  let title = document.getElementById('title-' + keyString);
+  title.innerText = "";
+  for(let i = 0; i<10; i++){
```
Leo2452, line 110: Is there a reason you iterate `10` times? It seems like you constructTripTitle per complete trip, so not sure why you need to loop. If I understood wrong (i.e. it actually loops through all trips in this stingle function), then you might want to change the `10` to the size of the trips saved for the user

renau06: so basically, this is just looping until the findPlace doesn't return an error. I wasn't sure which number to set it at bc sometimes it doesn't need to be delated and sometimes it needs to be delayed more than 1 second

byway/src/main/webapp/home.js
```diff
@@ -138,7 +163,7 @@ async function initMap(start, end, route, keyString) {
  * @param {Array} [waypoints] array of waypoint objects
  */
 async function setRoute(directionsService, directionsRenderer, start, end, waypoints){
-  for(let i = 0; i<20; i++){
+  for(let i = 0; i<30; i++){
```
zstewar1, line 166: You keep bumping this number up; if ~5 tires is not enough, you may need to use a longer delay.

byway/src/main/webapp/home.js
```diff
@@ -86,22 +87,46 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
+  title.innerText = titleString;
   title.href = configureTripKeyForPath(trip.keyString, "/routepage.html");
 }
 
+async function constructTripTitle(trip){
+  //let title = document.getElementById('title-' + keyString);
+  //title.innerText = "";
+  for(let i = 0; i<30; i++){
+      try{
+        const promises = trip.destinations.map(async (destination) => {
```
zstewar1, line 113: Oh, I see. You're starting the promises for every destination simultaneously, which is probably why you need to keep increasing the retry budget in `findPlace`. If you do these sequentially instead of simultaneously, you shouldn't need to keep adding more retries inside the `findPlace` function.

byway/src/main/webapp/home.js
```diff
@@ -44,7 +44,7 @@ function loadPastTrip(){
     tripIds.forEach(trip => {
       let isDestinationsMissing = trip.destinations.length == 0;
       let isInterestsMissing = trip.interests.length == 0;
-      (isDestinationsMissing|| isInterestsMissing) ? showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMissing) : showCompleteTrip(tripNum, trip);
+      (isDestinationsMissing|| isInterestsMissing) ? showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMissing) : showCompleteTrip(trip);
```
landondare, line 47: I know that this is not really part of this change, but can you please update this ternary operation into an if statement. Ternary operations are usually only used for assignment. An if statement would make this code more clear.

byway/src/main/webapp/home.js
```diff
@@ -86,22 +87,46 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
   container.append(pastTrip);
   initMap(trip.start, trip.start, trip.route, trip.keyString);
-  title.innerText = "Trip #" + tripNum;
+  let titleString = await constructTripTitle(trip);
+  title.innerText = titleString;
   title.href = configureTripKeyForPath(trip.keyString, "/routepage.html");
 }
 
+async function constructTripTitle(trip){
+  //let title = document.getElementById('title-' + keyString);
```
landondare, line 109: Make sure you remove these comments before submitting this PR

byway/src/main/webapp/home.js
landondare, line 113: Given the fact that you have to query the Maps API to get the name of places, It may be easier to store this as some string on the Trip object instead. It is not difficulty to add a new field to the Java object or datastore object. This way you wouldn't need this convoluted logic to ascertain the start and end names. 

The value of this string Title could be generated on the first step when you select the start and end points while you already have the string names. Or you could let the user enter one manually. Or auto generate a title based on the date, etc.

byway/src/main/webapp/home.js
renau06: done

byway/src/main/webapp/home.js
renau06: done

byway/src/main/webapp/home.js
renau06: ah okay i see, I changed it to sequentially and changed the loop to 5 tries

renau06: I updated the function to simply loop through and find the placeId of each destination and add to the title, so the logic is not too confusing anymore. I did think about creating a new field, but since alot of pages rely on creating trips, I did not want to mess up anything by adding a new field to the constructor. Also, since we are running out of time, I figured this would be the quickest way to get some sort of title, even though it would be cool to have the user enter their title manually

byway/src/main/webapp/home.js
renau06: fixed

byway/src/main/webapp/home.js
```diff
@@ -86,22 +92,49 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
+ * @param {Trip} trip
+ * @param {String} keyString trip's key as a string
+ */
+async function constructTripTitle(trip, keyString){
+  trip.destinations.forEach(async (destination) => {
+     for(let i = 0; i<5; i++){
+      try{
+        await delayPromise(1000);
+        let placeInfo = await findPlace(destination, placesService);
+        document.getElementById('title-'+ keyString).innerText += placeInfo.name + "|";
```
zstewar1, line 125: Since the caller awaits this function, instead of assigning from here, why not return the constructed name, and let the caller figure out where to put it.

byway/src/main/webapp/home.js
```diff
@@ -86,22 +92,49 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
 }
 
+
+/**
+ * Updates the title of specifice trip
+ * @param {Trip} trip
+ * @param {String} keyString trip's key as a string
+ */
+async function constructTripTitle(trip, keyString){
+  trip.destinations.forEach(async (destination) => {
```
zstewar1, line 120: The array `forEach` function does not wait for promises to complete any more than `map` does. If you want to wait you'd have to use `for( ... of .. )` with await.

byway/src/main/webapp/home.js
```diff
@@ -86,22 +92,49 @@ function showIncompleteTrip(tripNum, trip, isDestinationsMissing, isInterestsMis
...
+     for(let i = 0; i<5; i++){
+      try{
+        await delayPromise(1000);
```
zstewar1, line 123: You await on error, so you probably don't need to await before sending a query.

renau06: ah right! took it out

byway/src/main/webapp/home.js
renau06: fixed!

byway/src/main/webapp/home.js
renau06: fixed!

byway/src/main/webapp/home.js
```diff
@@ -44,7 +44,12 @@ function loadPastTrip(){
     tripIds.forEach(trip => {
```
zstewar1, line 44: Notice that up here you're also trying to load every trip simultaneously, so it might still be possible to hit your rate limits unexpectedly. If that becomes a problem you may want to await up here as well.

PR #51, opened 2020-08-31
Marker functionality

Leo2452:
Add double click feature to add a recommendation onto the route list. Add a toggle feature to remove / display recommendation markers


byway/src/main/webapp/recommendations.js
```diff
@@ -206,25 +217,32 @@ function placeMarker(place) {
...
- * to the google.maps object with true.
- * @param {Boolean} isNull indicates to remove markers by setting their map to null.
+ * Iterate through all markers and hide or show based on the 
+ * boolean tracking the toggle state. Only affect rec markers,
+ * not ones added to the route.
  */
-function showMarkers(isNull) {
-  let canvas = (isNull) ? null : map;
+function toggleRecMarkers() {
+  let canvas = (isHiding) ? null : map;
```
landondare, line 241: Nitt: would be useful to have a small comment on why this works. Why does setting the map to null hide the marker. Extra context in an inline comment would be nice.

byway/src/main/webapp/recommendations.js
```diff
@@ -30,8 +34,8 @@ const MAX_RECOMMENDATIONS = 1;
 // Holds regions google.maps.LatLng objects
 let regions = [];
 
-// holds markers as google.maps.Marker objects
-let markers = [];
+// Boolean to determine when to toggle markers
+let isHiding = true;
```
landondare, line 38: Nitt: would call this something slightly more specific and related to what is actually being Hidden. E.g. "waypoinsAreHidden" or "areMarkersHidden"

byway/src/main/webapp/recommendations.js

byway/src/main/webapp/recommendations.js
Leo2452: Fixed to be more specific. Also, changed some of the logic to track the true state of them (i.e. aren't initially hidden, only when toggled odd number of times - off by one state)

PR #52, opened 2020-08-31
Refactor js

j-vent:
Refactored routepage.js and recommendations.js to simplify a large, redundant request to the Directions API. Wrote helper functions to combine commonalities using promises instead of callbacks, and to separate out different function calls.

Modified and moved similar functions to base.js, which are used by the three core pages.

Added error message display to the routepage.

### DIFF
--- a/byway/src/main/webapp/base.js
+++ b/byway/src/main/webapp/base.js
@@ -1,6 +1,7 @@
 /** Script that contains functions shared and used across all pages */
 
-/* exported configureTripKeyForPath, getTripKeyFromUrl, setProgressBar, setupLogoutLink, findPlace */
+/* exported configureTripKeyForPath, getTripKeyFromUrl, setProgressBar,
+   setupLogoutLink, findPlace, getRouteForTrip, showErrorMessage */
 
 /** 
 * Sets Progress Bar to correct location based on the page number
@@ -92,3 +93,51 @@ function findPlace(placeId, placesService) {
   return result;
 }
 
+/**
+ * Class used to throw error in getDirections() with both message and status
+ */
+class MapStatusError extends Error {
+    constructor(message, status) {
+        super(message);
+        this.name = 'MapStatusError';
+        this.status = status;
+    }
+}
+
+/**
+ * Makes a request to the Directions API to return information about the route
+ * @param {DirectionsService} directionsService object that communicates with the GMaps API service
+ * @param {directionsRenderer} directionsRenderer object that renders display results on the map
+ * @param {String} start placeId as string
+ * @param {String} end placeId as string
+ * @param {Array} [waypoints] array of waypoint objects
+ * @return {Promise} result Directions route object with fields route, legs, etc.
+ */
+function getRouteForTrip(directionsService, directionsRenderer, start, end, waypoints) {
+   let request = {
+    origin:  {placeId : start},
+    destination: {placeId : end},
+    waypoints, 
+    travelMode: 'DRIVING',
+    optimizeWaypoints: true
+  };
+  const result = new Promise((resolve,reject) => {
+    directionsService.route(request, (result, status) => {
+      if (status == 'OK') {
+        directionsRenderer.setDirections(result);
+        resolve(result);
+      }
+      else {
+        reject(new MapStatusError("Could not calculate route from request.", status));
+      }
+    });
+  });
+  return result;
+}
+
+/** Displays error message to the user if a request fails.*/
+function showErrorMessage(errorMessage){
+  let msgContainer = document.getElementById("message-container");
+  msgContainer.style.visibility = 'visible';
+  msgContainer.innerText = errorMessage;
+} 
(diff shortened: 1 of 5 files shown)


byway/src/main/webapp/base.js
```diff
@@ -92,3 +93,51 @@ function findPlace(placeId, placesService) {
...
+        reject(new MapStatusError("Could not calculate route from request.", status));
+      }
+    });
+  });
+  return result;
+}
+
+/** Displays error message to the user if a request fails.*/
+function showErrorMessage(errorMessage){
+  let msgContainer = document.getElementById("message-container");
```
Leo2452, line 140: Since this is in the `base.js`, may not be guaranteed to work for all documents when using this script tag. This means all pages need a tag with that specific id. I think it'd be better to pass in the HTML element as a parameter and set it that way, rather than require all pages to have an element with that specific id.

byway/src/main/webapp/recommendations.js
```diff
@@ -125,7 +107,9 @@ async function loadRecommendations() {
       }
     }
   }
-  if(statuses.size !== 0) alertUser(statuses);
+   // String elements with status codes from placesService into an error message
+  let errorMessage =  "Showing limited results due to: " + Array.from(statuses).join() + ".";
```
Leo2452, line 111: small nit: spacing consistency

byway/src/main/webapp/routepage.js
```diff
@@ -72,25 +79,34 @@ function initMap() {
...
+ */
+async function getRecommendations(){
+  try{
+    let result = await getRouteForTrip(directionsService, directionsRenderer, start.place_id, end.place_id, 
+      route.map(waypoint => ({location: waypoint.geometry.location})));
+    findRegions(result);
+    loadRecommendations();
+  } catch (error) {
+      showErrorMessage(error);
+
```
Leo2452, line 108: small nit: extra newline here

byway/src/main/webapp/recommendations.js
j-vent: Changed!

byway/src/main/webapp/routepage.js
j-vent: Changed!

byway/src/main/webapp/base.js
```diff
@@ -92,3 +93,51 @@ function findPlace(placeId, placesService) {
...
+    origin:  {placeId : start},
+    destination: {placeId : end},
+    waypoints, 
+    travelMode: 'DRIVING',
+    optimizeWaypoints: true
+  };
+  const result = new Promise((resolve,reject) => {
+    directionsService.route(request, (result, status) => {
+      if (status == 'OK') {
+        directionsRenderer.setDirections(result);
```
zstewar1, line 127: Try to keep the operation of converting `directionService.route` into a promise from what you want to do with the results. 

I'd split this in 3 functions:

1. Wraps `directionsService.route` and makes it return a promise. something like `computeRoute(request, directionsService)`
2. Handles getting the route for the current trip by calling that function. like `computeRouteForTrip(directionsService, start, end, waypoints)`, which just takes care of building the request and calling `computeRoute`.
3. Chaining off of or awaiting `computeRouteForTrip` to call `directionsRenderer.setDirections`

byway/src/main/webapp/base.js
j-vent: I spoke to Rena about also adding errors to the pages she worked on, and suggested message-container be a common element in the base.css. So we do have the message-container elements coming up in another PR, but IDs are supposed to be unique so I'm not sure if this the best way to do it either. Will wait for Zach or Landon to comment.

byway/src/main/webapp/base.js
j-vent: Thanks! I've separated getRouteForTrip() into two separate functions for cleaner structure and reuse.

byway/src/main/webapp/routepage.js
```diff
@@ -72,39 +79,36 @@ function initMap() {
...
-      window.alert("Could not calculate route due to: " + status);
-    }
+/** Uses the route from the directionsService request to update 
+ *  ordering of stops on the route panel, and distance and time of roadtrip
+ */
+async function updatePageInfo(){
+  try{
+    let result = await computeRouteForTrip(directionsService, start.place_id, end.place_id, 
+      route.map(waypoint => ({location: waypoint.geometry.location})));
+      directionsRenderer.setDirections(result);
```
zstewar1, line 89: Nit: unindent this line to match `let result = ...`; since it's indented the same as the continuation of the above line, it looks like its also a continuation of the previous command rather than a new line.

zstewar1 (dismissed):

byway/src/main/webapp/routepage.js
j-vent: Did not catch that indent, fixed!

